PR #27, opened 2021-09-06
Add partial support for Ubisoft Connect

cftad:
Raised to partially address #15 

- Further investigation is required for the obtaining of game artwork. 
- Adding the games to the library has been omitted due to the recent filter changes. 

## Omitted Example 
`GameGridPage.xaml.cs`
```cs
async Task LoadGamesAsync()
{
    var steamLibrary = new SteamLibrary();
    var steamGames = await steamLibrary.ListGamesAsync();

    var ubiLibrary = new UbisoftLibrary();
    var ubiGames = await ubiLibrary.ListGamesAsync();

    List<Game> masterLibrary = new List<Game>();
    masterLibrary.AddRange(steamGames);
    masterLibrary.AddRange(ubiGames);

    masterLibrary.Sort();

    DispatcherQueue.TryEnqueue(() => {
    foreach (var game in masterLibrary)
    {
        Games.Add(game);
    }
    });
}
```

PR #33, opened 2021-09-12
Performance improvements and UI tweaks

MorGuux:
I've changed a few things since release v0.9.3.0.

* Merged Ubisoft support from #27
* Games now appear when they're loaded. With a blurred background, this now lets the consumer see games load in as they're found.
* Major overhaul to DLSS DLL search function, now using a searcher class that instantiates a new searcher for each HDD/SSD on the user's system. Stops searching once it finds the first match (although this needs to change to detect multiple versions of DLSS for a game)
* Migrated the string version to a Version class. This can simplify comparing two versions, and some code improvements.
* Changed the DLL download progress bar to use progress reports (updates after each percent, not every 10%)
* Added a "Found x games" counter on game loading, can be nice to illustrate things are happening.
* Updated the steam library search function by using an external lib [Sledge-Formats](https://github.com/LogicAndTrick/sledge-formats). Much faster searching for library folders and app manifests than previous implementation.
* Tweaked headerimage stretch to accomodate other aspect ratios in the future
* Tweaked game GridView margins to fix the box position on hover.

### DIFF
--- a/DLSS Swapper/DLSS Swapper/DLSS Swapper.csproj
+++ b/DLSS Swapper/DLSS Swapper/DLSS Swapper.csproj
@@ -27,6 +27,7 @@
     <PackageReference Include="Microsoft.ProjectReunion.Foundation" Version="0.8.3" />
     <PackageReference Include="Microsoft.ProjectReunion.WinUI" Version="0.8.3" />
     <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" />
+    <PackageReference Include="Sledge.Formats" Version="1.0.2" />
     <Manifest Include="$(ApplicationManifest)" />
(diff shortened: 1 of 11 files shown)


KedarWolf:
Oh, can't wait for foe the aspect ratio tweak. I used a 3840x1080 screen that DLLS 1.0 doesn't support. :)

MorGuux:
I think you're misunderstanding the software. It is a manager for the DLSS dlls, that lets you swap to different versions created by NVidia. It does not change how DLSS functions/works, just which version of DLSS the game uses.

KedarWolf:
Oh, I thought the tweak might enable support for screen resolutions that DLSS 1.0 doesn't support natively. :(

PR #73, opened 2022-01-22
Fix for infinite loop caused by symlinks cross references

tamodolo:
Fixed the infinite loop that happen when cross references are made with symlinks. Occur when using injection mods like Special K.

Thanks to beeradmoore that helped a lot to narrow the problem and helped with the C# code.


beeradmoore:
Awesome. Looks good.

How'd you like to be attributed in the app? (I'll add contributions to the repo at some point as well)
Mine is 
> Brad Moore ( [@beeradmoore](https://twitter.com/beeradmoore) )

with the link going to my Twitter. Can add a single link to a social, or github account, or none. We don't really have a standard 😂

You can even have just minimal 
> tamodolo 

with no links. Whatever you're comfortable with.

tamodolo:
I think just the github is fine.

PR #117, opened 2023-02-13
Import Microsoft.PowerShell.Archive and force Compress-Archive

thiagodpaz:
Fixes 'CouldNotAutoloadMatchingModule' error on ./package/build_Portable.cmd

image

### DIFF
--- a/package/build_Portable.cmd
+++ b/package/build_Portable.cmd
@@ -30,7 +30,7 @@ echo Zipping app
 echo ################################
 echo.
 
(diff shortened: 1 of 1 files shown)


beeradmoore (approved):

beeradmoore:
Thanks for the PR. I wasn't getting these errors originally. I can remember pwsh scripts being a pain in the past because of execution policy, I assume the first part of this avoids that.

But I tested it and it still works so sent it through. Thanks again!

PR #151, opened 2023-08-22
Custom Directories Feature

BmKo:
Had a go at adding custom directories support, never worked on a UWP app before so hopefully I went about adding to the settings page correctly.

* Added multiple custom directories support #8 
  * Detects game folders within tracked directories
  * Asynchronously checks each tracked directory stored in `settings.json`
  * Does not support images 
    * Could be done using [IGDB](https://www.igdb.com/api) although requires API key
    * Otherwise detect custom image file by filename in game directory
* Switched `App.xaml` to only use `Windows.UI` for colours (I had compatibility issues building without it)

## Demo
DLSS_Swapper_zZkEvx1V9w

### DIFF
--- a/src/App.xaml
+++ b/src/App.xaml
@@ -2,7 +2,8 @@
     x:Class="DLSS_Swapper.App"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:local="using:DLSS_Swapper">
+    xmlns:local="using:DLSS_Swapper"
+    xmlns:ui="using:Windows.UI">
     <Application.Resources>
         <ResourceDictionary>
             <ResourceDictionary.MergedDictionaries>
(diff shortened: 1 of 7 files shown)


beeradmoore:
Hey @BmKo , thanks for the PR!

I may be able to do this from my end (will try later), but can you change the target branch to be `feature/custom-directories` instead of `main`?

Of all the thinking I have done about doing this, had not crossed my mind that Settings needs UI to add folders. I was too focused on the new UI for game details to replace the DLSS dropdown 😅 (I'll draw up some pics and get feedback in #8 soon)

I am not 100% sure if game directories will work as intended like you have done here.
EG:
```
C:\Games\
├── GoG
│   └── GameA
├── Freebies
│   └── GameB
└── OtherPath
    └── SubPath
         └── GameC
```
(I have not checked the full source yet so sorry if my thoughts here are incorrect) In this example if you added `C:\Games\` I don't think it will pickup `GameA`, `GameB` and `GameC` but instead `GoG`, `Freebies`, `OtherPath`.

My thoughts were that a user would use the UI you made in settings and then explicitly add `GameA`, `GameB` and `GameC` directories. May be possible to have a "Add game" and "Add games in path" as two seperate buttons and if they select the second option they need to make sure their folders are right?

From the new DLSS dropdown it will actually have more information about the game. On the right I am putting the cover, clicking this his is where you can select and change the cover art for any game. This is also the thing where the user can add their own game cover that they saved from Google/IGDB/etc for those games that are manually added.

Top will have game title, this can be editable by the user if they want to rename a game (eg. World of Warcraft to just WoW because 🤷‍♂️). Middle area will have some details like game path, open in explorer, dlss version, dlss frame generation (AKA DLSS 3) dropdown, and then this is where also XeSS and FSR drop-down would go.

Bottom so far I have one icon which is a notes icon which allows the user the enter custom notes about a game that just get saved locally so they can remind themselves of "oh shoot, don't use DLSS 3.0.1 here, game crashes".

BmKo:
No worries wanted to use it myself, so gave it a go. Also it looks like you will have to create the branch on the repo for me to reference it.

> I am not 100% sure if game directories will work as intended like you have done here. EG:
> 
> ```
> C:\Games\
> ├── GoG
> │   └── GameA
> ├── Freebies
> │   └── GameB
> └── OtherPath
>     └── SubPath
>          └── GameC
> ```
> 
> (I have not checked the full source yet so sorry if my thoughts here are incorrect) In this example if you added `C:\Games\` I don't think it will pickup `GameA`, `GameB` and `GameC` but instead `GoG`, `Freebies`, `OtherPath`.

Yep I did just set it up to use the sub directories as game names, so in the example above you would have to add each `GoG`, `Freebies`, `OtherPath\SubPath` to detect the games and their names. Although if it were to just get the folder that `nvngx_dlss.dll` resides in it could be a sub directory of the game like how Unreal Engine games are structured `\Engine\Plugins\Runtime\Nvidia\DLSS\Binaries\ThirdParty\Win64`.

> My thoughts were that a user would use the UI you made in settings and then explicitly add GameA, GameB and GameC directories. May be possible to have a "Add game" and "Add games in path" as two seperate buttons and if they select the second option they need to make sure their folders are right and it would have the same result as hitting the first button multiple times?

I think for people like me who keep their games somewhat organized the add directory button would save a bunch of time, it just might need an error message if a sub directory contains multiple `nvngx_dlss.dll` files. But a "find games" button would still work pretty much the same. Add game and find games would just have to filter out non-dlss games, so the game list doesn't get cluttered.

> Other than that, not sure what the deal is with the Windows.UI colours thing. I'll see if I can dig up any info on why that would need to be changed.

Yea I use Rider as my IDE and that was the solution it recommended for the error as there were multiple choices for `Color` (there was also `System.drawing.Color` but someone on [StackOverflow](https://stackoverflow.com/a/55683085) mentioned UWP is not compatible with it, they also suggest `SharpDX` but `Windows.UI` does the job).

beeradmoore:
> Although if it were to just get the folder that nvngx_dlss.dll resides in it

Unfortunately it doesn't always exist in the root path so things will be off. Same reason we can't just hope every game is unreal and use that path you provided.

> ... it just might need an error message if a sub directory contains multiple nvngx_dlss.dll files

Unfortunately I have come across games that have two -_- 

Would be interested to know what GeForce Experience is doing to know the names better, unless they just have the resources and contacts to be able to fingerprint every game path correctly.

I have Rider but I use Visual Studio. I'll see if I can have a look why it's doing that, although the fix you did might just be the fix in general.

charon166 (approved):

beeradmoore:
Hey @BmKo ,
RE: Rider issue. I was able to replicate. It didn't appear to prevent building even though it was reporting problems. I have filed a issue with [JetBrains YouTrack system](https://youtrack.jetbrains.com/issue/RIDER-101258/) as I believe this is an IDE issue not a code issue. I will revert the changes you did to `App.xaml` in the mean time. 

Hopefully it is an easy fix for them to get on top of. If I get any updates of the above ticket I'll let you know here.

Thanks for the PR, big step forwards for adding custom games and I am looking forward to building on top of this.

PR #161, opened 2023-09-24
Write DisplayVersion, Publisher and DisplayIcon to ARP entry

SpecterShell:
This PR fills these two empty fields and replace the default icon with the program's for the future installers. You can choose another name for the Publisher field.

Before:

image

After:

image


beeradmoore (approved):

beeradmoore:
Thanks heaps for that @SpecterShell. I had always wondered why the icon never showed, didn't even notice version number did not show.

PR #223, opened 2024-08-21
Fix app crash when no internet connection is available

### LINKED ISSUE #92: [BUG] DLSS Swapper crashes if you don't have internet connectivity.
**Describe the bug**
Crash when not online.

**To Reproduce**
1. Disconnect internet. 
2. Open DLSS Swapper.

**Expected behavior**
App works but prevents me downloading new DLSS Versions or checking for updates.

**Screenshots**
N/A

**Additional context**
This came to light while investigating #84, but was not directly related to it.
Also with the new local thumbnail images being used it helps using DLSS Swapper in offline mode.

Zagrthos:
### Summary
This resolves #92.

### Detailed Changes
Added a try catch around the `HttpRequestMessage`.
If an `HttpRequestException` occurs it will return and display an error message to the user about checking the internet connection.

I also had to remove the `ConfigureAwait(false)` addition at the `HttpClient#GetAsync` statement because it would've crashed the app again when the connection is back (Don't ask me why but the bug was gone after I removed it!).

### Notes
I also made small formatting changes to the methods (nothing breaking tho, this PR is only intended to fix the issue).

### DIFF
--- a/src/Data/DLSSRecord.cs
+++ b/src/Data/DLSSRecord.cs
@@ -156,15 +156,30 @@ namespace DLSS_Swapper.Data
 
             _cancellationTokenSource?.Cancel();
 
-            LocalRecord.IsDownloading = true;
-            LocalRecord.DownloadProgress = 0;
-            LocalRecord.HasDownloadError = false;
-            LocalRecord.DownloadErrorMessage = String.Empty;
-            NotifyPropertyChanged("LocalRecord");
-
             _cancellationTokenSource = new CancellationTokenSource();
             var cancellationToken = _cancellationTokenSource.Token;
-            var response = await App.CurrentApp.HttpClient.GetAsync(DownloadUrl, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+
+            HttpResponseMessage response;
+            try
+            {
+                response = await App.CurrentApp.HttpClient.GetAsync(DownloadUrl, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+
+            LocalRecord.IsDownloading = true;
+            LocalRecord.DownloadProgress = 0;
+                NotifyPropertyChanged("LocalRecord");
+            }
+            catch (HttpRequestException ex)
+            {
+                Logger.Error(ex.Message);
+
+                LocalRecord.IsDownloading = false;
+                LocalRecord.HasDownloadError = true;
+                LocalRecord.DownloadErrorMessage = "Could not download DLSS. Please check your internet connection!";
+            NotifyPropertyChanged("LocalRecord");
+
+                return (false, "Could not download DLSS. Please check your internet connection!", false);
+            }
+            
             if (response.StatusCode != System.Net.HttpStatusCode.OK)
             {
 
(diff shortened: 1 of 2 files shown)


beeradmoore:
Looks good, as with the other one I'll test on the weekend. 

I like the change of using `string.` instead of `String.`, I've also started doing that in day to day work. 

I hope these two things aren't too nitpickey.

I prefer the use of `IsNullOrEmpty` over `IsNullOrWhitespace` as the intent is more clear, that I am either expecting `null` or `string.Empty`. It will of course catch `" "` but that shouldn't be coming through, and if it did should we also check for `"a"` as that would not be ideal. Is it ok to swap that back?

As for the other formatting change, is there any reason why you went with `is` over `==`, or is that just a personal style choice?

eg.
```csharp
if (result.Success == false && result.Cancelled == false)
```

vs. 

```csharp
if (result.Success is false && result.Cancelled is false)
```

I find the first more readable, but ironically I find myself using `is` for nulls like `if (result is null)`. And a lot when like `if (result is string resultString)` when needed.

Zagrthos:
> I prefer the use of IsNullOrEmpty over IsNullOrWhitespace [...]. Is it ok to swap that back?

Since we got trained with that in apprenticeship (and it's also the Microsoft default in each tutorial they publish) I personally went with `IsNullOrWhitespace` as my default checking too. We could replace that with a `Regex` (oh hell no) which checks if the string is a valid Uri. Otherwise we could just try to construct a new `Uri` object with `Uri#TryCreate` and see if it's possible or not.

> As for the other formatting change, is there any reason why you went with is over ==, or is that just a personal style choice?

Personal preference as I started to use `is` waaay more often than `==` because it's mainly safe from overloading. You're able to overload the default `==` operators with different functions but you're unable to do that with `is`. See [here](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#null-checks).

Although I'm comfy with changing both back to the original conditionals. But after all you're the maintainer 🙂.

beeradmoore:
> oh hell no

😂

I think validating URL is going above and beyond and not required. If it was a user entered URL, sure 100% yes, but it isnt. So long as its not null or empty we should be good. Curious that they are checking whitespace as well by default. IMO I had avoided it because if I am not expecting whitespace why check for it and then my code will go faster. But I am 110% sure it is negligable, especially in the places we use it :P 

Thinking about it `IsNullOrWhitespace` is another check I'd do if it was user input incase they accidentally hit space or something. Interesting choice by MS though.

> because it's mainly safe from overloading

Makes sense, and that is good point. I don't think I've ever ever overloaded it, but it has got me in the past when I have overloaded `Equals` and I went on to use `==` to find it doesn't work.

I will keep this in mind for when I do code in future, for now is it ok to revert those back to `==`?

As for `if (!result.Success)`, I have poor vision and often can't see that `!` hiding in there so I'll always write it out in the full `if (result.Success == false)` for readability.

Zagrthos:
Alright, updated the PR. Changed back to `IsNullOrEmpty` as well as did another formatting pass like in the other one I did with removing some unneeded `using` directives. 🙂

beeradmoore (approved):

Issue #42, opened 2021-10-06
[Feature request] System notification for DLSS updates


beeradmoore:
**Is your feature request related to a problem? Please describe.**
When a new version of DLSS is updated we don't know about it unless we manually check.

**Describe the solution you'd like**
If its possible to ding the messaging center I think we should.

**Additional context**
Needs to be user togglable.

MorGuux:
https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=builder-syntax

OzzyD1:
On top of that, I think a notification if a game update rolled back a DLSS version would be very useful.

Zagrthos:
I see of different ways to solve this.
1. Adding another JSON file containing an array of a model from each identified game with properties regarding the DLSS dll (timestamp, version, etc)
2. Using [EF Core](https://github.com/dotnet/efcore) we could add a local SQLite Database (maybe too much but the use-case is there)


Either way we can compare the info against the newly retrieved data when scanning for games and push out a notification or not.

However if the intention was to do that in background (like scanning each few minutes the paths) I think that's not possible using a WASDK App. Correct me if I'm wrong.

beeradmoore:
The new in progress branch has SQLite (in the form of sqlite-net). 

The DB was because we needed an easier way for system to store manually added games as well as a bunch of new info applied to all games (custom images, notes, etc)

A side effect of this is we will be able to load games from cache while updating them in the background followed by looking for new games. Making the time from launch to modifying records faster.

We can certainly keep track of "last swapped" and alert the user (either system notifications, or in app notification list).

If it went down assume the game updated and replaced it. If it went up assume the game updated with a new version.

On this topic (I'll make an issue for it soon) I was also thinking that DLSS Swapper should be able to import/auto-import DLSS/FG/RR versions into its DLSS library. The idea being that game XYZ updates and has a new unseen version of DLSS Swapper, that is then imported and ready to be used in any game. In theory none of that should require an internet connection, it could do it all offline.

beeradmoore:
I have went down a rabbit hole and its gross.

I can get notifications to work, however they don't work when running as admin. Because running as admin is becoming required for some instances I needed another solution. I prototyped up a C++ app (because it will be nice and small compared to C# app) which should be able to be invoked from DLSS Swapper as a non-admin user solving the issue of notifications for people running as admin.

[...]

PR #224, opened 2024-08-21
Feature: Add setting to hide non downloaded versions in selector

### LINKED ISSUE #208: [Feature request] option to hide non-downloaded DLSS versions
What I would like to see is a setting where, if enabled, the list of DLSS versions to load to the game gets restricted to the downloaded versions. There could be situations where people are picky about specific versions and do not care about the others. In my case, it can simplify the options list a good deal.

Zagrthos:
### Summary
This resolves #208.

### Detailed changes
Added a new "Choose DLLs To Hide" button to the settings page which defaults to nothing selected.
The new settings are respected in the `DLSSPickerControlModel` and filter out the non-downloaded versions which ARE NOT the currently found version of the Game. So there is always at least one version visible for the user.

### Notes
Everything was tested by myself for functionality.


beeradmoore:
Thanks for this PR, I'll take a look at it on the weekend.

beeradmoore:
I think there is an issue with selecting DLSS verions in the dropdown.

I enabled the feature to hide DLSS versions that are not downloaded

Screenshot 2024-08-25 142922

This says I can install v3.6.0.0, but if I turn the feature off there is no DLSS v3.6.0.0 shown.

Checking the library I do have v3.6.0.0 downloaded.

Zagrthos:
Seems like my tests weren't that excessive that I thought I covered all cases.

Pushed a fix and now it seems to work.

beeradmoore:
I will test this when I get to testing the other one again

Zagrthos:
> The wording in settings may change in the future, but its good enough for now. I think non-downloaded works better than not downloaded. But I am not 100% sure on if its the correct words to use.

I choose not downloaded because it implies that this version has not yet been downloaded by the user. Non-downloaded would be something do describe it as something different (from my POV).

> The only other thing is I wonder what the use case should be for a game with DLSS versoin not in the downloaded list. I think this looks a little confusing (especially if the panel covers the game)

Here we can solve this by probably copying this one to the DLSS Swapper download directory where the other files lie, then it's get's detected as well and is shown inside the list. Would also fix the follow-up comment

> But also what we have currently is an option saying to download the version that is already installed

> Maybe based on that other idea of improting DLSS from games into the library, when a user wants to change a DLSS version from one that does not exist it tells them this will be automatically imported or something.

I probably just described this in other words, lol.


> But then it still has the issue of if you go from 3.5.10 (not in library) to 3.6 to 3.7 that now 3.5.10 is gone. The button to reset also won't work as that would only go back to 3.6.

We could do 3 fields in the DB for this.
1. Initial version with what the game was delivered (game was found the first time)
2. Swapped version through user
3. Backup version is the last version which was swapped by the user before changing to the new one

beeradmoore:
> Non-downloaded would be something do describe it

Yep, that makes sense. I was also thinking "local" but it also doesn't feel right. 

> Here we can solve this by probably copying this one to the DLSS Swapper download...

Seems copying to local automatically (without even alerting users) may solve a bunch of problems. Unsure if it introduces new problems of users filling up their downloads list if they don't want all versions there. We could add a hide button in the library. Also only enable delete if the DLSS version is not an initial or current version (see below) for any game. 

There is a seperate issue for that here, 
https://github.com/beeradmoore/dlss-swapper/issues/227

> We could do 3 fields in the DB for this.

We would also need to handle new version of DLSS detected in game folder and update that as "initial". We could then just do 2, initial and current. On launch if initial is not set we set it to actual from disk. If current is not actual from disk we update initial to be current. Setting initial would also copy initial to be the backed up version on disk.

Detecting of current changing would also be the point we alert for [this](https://github.com/beeradmoore/dlss-swapper/issues/42#issuecomment-2320337586) comment.

beeradmoore:
I finally got all the code merged into my branch, and updated with WindowsAppSDK 1.6. This is what I was thinking of with the new UI for picking a DLSS version. I did not have version picker added here yet. I will do so soon with that combobox.

image

Zagrthos:
Now that's a good looking one. Good job! 🙂
And finally having updated to the latest version, haha. Looked forward to that!

Zagrthos:
I'll try and gonna check if I can merge main _somehow_ into this branch. The changes are massive and I already got 5 conflicts for it 😄

beeradmoore:
Yeah sorry about that. Although I think things will be easier if you branch from development and start again.

You should be able to bring the settings across easy, and in the new `DLLPickerControlModel` is where you can check settings and remove from the [new lists we make here](https://github.com/beeradmoore/dlss-swapper/blob/main/src/UserControls/DLLPickerControlModel.cs#L60) ;)

Zagrthos:
> Yeah sorry about that. Quite a few things have changed. Although I think things will be easier if you branch from development and start again.
> 
> You should be able to bring the settings across easy, and in the new `DLLPickerControlModel` is where you can check settings and remove from the [new lists we make here](https://github.com/beeradmoore/dlss-swapper/blob/main/src/UserControls/DLLPickerControlModel.cs#L60) ;)

Don't worry, I understood the new code pretty fast. 🙂 And I like the new MVVM approach.

I force pushed back to the commit where I branched from main, then merged main and merged my (new) changes back inside.

PR is Ready for Review again.

beeradmoore:
Thanks for the fast fix! I will re-visit this after v1.1.5 is out (hopefully tonight)

Ever since I had started doing bulk changes to how a lot of pages and code works I keep seeing the "Pull Requests (1)" up the top and thinking oh boy I hope this one can be fixed

Zagrthos:
Haha no worries!
I am honest I actually forgot I still had this open PR here because somehow the project went fully into the back of my head. 😄

Great to see you didn't abandoned the app tho!

beeradmoore:
Sorry, we got some conflicts in settings after the recent changes

src/Pages/SettingsPage.xaml
```diff
@@ -103,6 +103,30 @@
...
 
+                <TextBlock
+                    Style="{StaticResource SubtitleTextBlockStyle}"
+                    Margin="0,40,0,0"
+                    Text="Hide Not Downloaded DLL Versions" />
+                <StackPanel Margin="0,10,0,0">
+                    <DropDownButton Content="Choose DLLs To Hide">
+                        <DropDownButton.Flyout>
+                            <MenuFlyout Placement="Bottom" ShowMode="Standard">
+                                <ToggleMenuFlyoutItem x:Name="HideDlss" Text="DLSS" IsChecked="{x:Bind ViewModel.HideNonDownloadedDlssDlls, Mode=TwoWay}" />
```
beeradmoore, line 114: For these, I think we can just have 1 option of "hide not downloaded DLL versions" instead of one of each. Unless people would want to hide FSR and not hide DLSS FG.

Zagrthos: My idea was to give people maximum flexibility but I found out myself that these are a little too much options. I'm shrinking it down.

Zagrthos: Changed to be one single setting.

Zagrthos:
> Sorry, we got some conflicts in settings after the recent changes

Conflicts are solved now.

beeradmoore:
Perfect, I'll have a look after work today

beeradmoore (approved):

beeradmoore:
There was something about "hide non downloaded" that just didnt feel right, I don't know what it is. I renamed it to "only show downloaded". Same thing, but removes the negation in the middle. Seems to fit better.

image

I also found situations where you could just not have any DLLs in the list so I added an info panel

Screenshot 2025-02-09 134001

Zagrthos:
Thanks for refining! 🙂

The info box indeed is a good thing which I probably missed. Thought about it tho but since in my tests there was at least one DLL in each game (which supported any tech) it never happened...

Issue #205, opened 2024-02-24
[BUG] DLSS Swapper first launch runs as administrator


beeradmoore:
**Describe the bug**
When installing from the installer you can select to launch application. Because the installer is running as admin the process will also run as admin. 

**To Reproduce**
1. Install DLSS Swapper
2. Launch from end of installer
3. Look at [task manager details](https://superuser.com/a/1213106/259386) tab to see if the process is running elevated
4. Close DLSS Swapper
5. Open DLSS Swapper from start menu
6. Observe that is is no longer running elevated

**Expected behavior**
App does not run as admin

**Screenshots**
N/A

**Additional context**
Quick fix is to remove the launch from installer. Other answers may be to look at runas and see if we can make runas run as invoker or current user.

beeradmoore:
Fixed as we simply don't launch from the installer anymore

Issue #91, opened 2022-10-09
[Feature request] Admin rights handling when swapping DLSS in Xbox app games


revanmj:
**Is your feature request related to a problem? Please describe.**
As discussed in #2, it seems that in some cases Xbox app game folders require admin rights for file operations inside them and DLSS Swapper currently doesn't handle it well (it just fails with generic error suggesting repairing game files, but works fine if whole app is launched as admin).

**Describe the solution you'd like**
Maybe in case of a fail retry swapping with admin rights just for Xbox app games or add additional error dialog with such option?

**Additional context**
My guess is that on PCs updated to Win10/11 from older versions, partitions or permissions may have been handled differently than on clean installs of those systems which leads to this (as I don't have this problem on my laptop with clean install of Win11, only on PC which was not formatted since Vista, only upgraded).

beeradmoore:
Bringing across my comment from the other thread.

==========
Interesting... 

For me on both of my systems (neither of which I have done funky things with WindowsApps folder) I can write to the game folder while using my normal accounts. I can't view the WindowsApps folder itself, so if I am in a game folder and attempt to go up a directory it says no.

For reference of what I have going on here - one system is Win11 where I am logged in with my MS account. The other system is Win10 where I am logged in as a local account. Both systems I am logged into the same MS account for Xbox app.

This new feature works by using the .GamingRoot files which are stored on the root of every drive. It contains the location of your Xbox games (what is the right word for this, Xbox app games?) are installed on that drive. It then lists these directories and compares against a list of every package installed on your computer to say. It then keeps a reference to the game location in WindowsApp folder which as I said above, the individual folders (at least these ones) are fully writable to me. The files added here also instantly add to the other folder (C:\XboxGames\ for me)

I am in Xbox app Insider and I have used the [Advanced Installation and Management Features](https://news.xbox.com/en-us/2021/11/18/advanced-installation-and-management-features-available-for-insiders-through-the-xbox-app-for-windows/) for the Xbox app, but the games I had tested with were not "advanced", they just let me install the game location.

My test game was The Ascent from Game Pass.
==========

Some additional info to add is both PCs I have I did install that particular OS from scratch. I will look into checking permission and prompting for admin access where required though.

revanmj:
Regarding the folder, I used D:\Program Files\Xbox Games\ in case it could also matter (I have 3 physical SSDs with one partition on each)

beeradmoore:
[...]

PR #225, opened 2024-08-22
Add run as administrator functionality

Zagrthos:
### Summary
This resolves an internal TODO inside the codebase to implement a proper working "Run as administrator" button dialog.

### Detailed Changes
The methods `IsRunningAdmin` and `RestartAsAdmin` are implemented inside the App.xaml.cs file. This was made because the old methods resided there and I just extended/replaced them.

The button dialogs in `GameGridPage.xaml.cs` were modified to use the newly introduced checks to work with. If a user chooses to run the app as admin it will be closed while a new one will be opened.

Also I added a new `Infobar` at the top of each page indicating if the app runs with administrative permissions.
image

### Notes
Additionally to the above changes I slightly modified the header layout of the Settings page to be more in line with how the other pages look.
This was necessary as otherwise the info bar would have looked absurdly crazy.
image

And while I was busy with that I also unified the XAML code of the `LibraryPage.xaml` file with how the other two look now.

### Edit
I also deleted the unused `BoolToVisibleConverter.cs` since inside the codebase only the CommunityToolkit approach is used.


beeradmoore:
Adding for additional reference:
- #91, needing to run as admin was first discussed here, this is where that initial code you found came from
- #205, not directly related, but related to run as admin, if you launch DLSS Swapper directly from the installer it will run as admin as the installer runs as admin. I like the idea of the banners reflecting this.

Ideally running as admin isn't required, but if it is there and it works then that's great. So long as we don't nuke a users system accidentally 😬

I'll jump into the PR more on the weekend.

Those banners look great. Good job!

Good call for removing converter that isn't used.

Just a FYI, in the branch `feature/custom-directories` is where I am working on the manually add games feature. This will change the `GamesGrid` page, including the DLSS picker dropdown.  With the PRs you are doing I wanted to give you the heads up to not adjust that popup as it will be fully replaced and I wouldn't want you to waste your time on something that gets removed.

After review I'll bring your work into `main` and then `main` back into my branches and fix up any conflicts there.

Zagrthos:
> So long as we don't nuke a users system accidentally

If only the directory where the game is gets modified I'm almost sure it wouldn't nuke the system - but I haven't implemented something that would remove System32 😄 

> This will change the GamesGrid page, including the DLSS picker dropdown.

I think besides changing the logic behind the two pages/controls (and adding the warning banners) I didn't touched them much.

Good heads-up tho with the info, I probably would've started getting rid of more issues which maybe involve the `GameGridPage`.

beeradmoore:
Trying to test this and I am unsure if my test setup is correct. 

I went to my Cyberpunk 2077 folder and I changed the permissions so my PC user can't write to this folder, 

Screenshot 2024-08-25 145537

But an "administrators" can, 

Screenshot 2024-08-25 145528

My account is an administrator account, but I assume "administrators" is when you run something as admin. Trying to change a dll I get the prompt to relaunch as admin. After I relaunch I get the error of not being able to write to the target directory. From task manger I confirmed that the relaunched instance is running as elevated (which is the same as right mouse click > run as admin).

It may not be working as intended because the permissions I applied are not correct to replicate the problem we are trying to solve.

I am downloading The Ascent from the Windows Store as it is what I had a problem with before, but I am pretty sure it is installing with the same mechanism as the Xbox app does which would be a valid install.

Do you have any recomendations on how to test this?



Aside from that the only other thing I am not 100% sure on is the text in the admin banner. I don't like how it says "Administrative Privileges" twice so close to each other. I tried changing the title to "Warning", its better, but it still isn't great. Mabye just not having a title at all? Any thoughts on that?

image

I was also thinking about if we needed a "disable running as administrator banners" in settings but I think the value it adds is tiny compared to starting to clutter the settings page it would do for every other user. So probably not.

Zagrthos:
> I was also thinking about if we needed a "disable running as administrator banners" in settings but I think the value it adds is tiny compared to starting to clutter the settings page it would do for every other user. So probably not.

We can flip the `IsClosable` switch. Then users are able to click on the "X" and it vanishes. My initial intent was to "force" showing this as administrative privileges are somewhat dangerous.

> Aside from that the only other thing I am not 100% sure on is the text in the admin banner. I don't like how it says "Administrative Privileges" twice so close to each other. I tried changing the title to "Warning", its better, but it still isn't great. Mabye just not having a title at all? Any thoughts on that?

Changed this bugged me in the end too but I wanted to wait for feedback. 😄

> My account is an administrator account, but I assume "administrators" is when you run something as admin.

`Administrators` is actually the built-in windows group, same as for `Users`. But I tested this right now with a clean `User` (no additional perms and denied writing perms on a Steam folder). I get the same error now. Currently investigating.

Zagrthos:
Okay I got to a point.
If you have denied writing perms on the `Users` group you are unable to test this without an explicit account which is only inside the `Administrators` group. Your default user account (which is created after the successful installation of windows) is in both groups (`Users` and `Administrators`). And because the denial of a permission weights higher than granting a perm you are unable to do that.

beeradmoore:
> We can flip the IsClosable switch. Then users are able to click on the "X" and it vanishes. My initial intent was to "force" showing this as administrative privileges are somewhat dangerous.

I was thinking to disable it completely and never show again. I think lets keep it as is without the "X" and if people don't like it (realistically most people will never see it) we can look at what would be the best way to handle it then.

I'll try get to re-test this by the weekend, but if not it will be on the weekend.

beeradmoore:
I did not get time this weekend, I will try ASAP though

beeradmoore:
> Edit 2: I got a test case now and verified it's working correctly.

The extra thing I had to do here was remove the permissions for `Authenticated Users`. I was able to test, see the failure, relauch, and then successfully change DLSS. 

Thanks for your work on this one.

PR #348, opened 2025-02-02
New bug template

Saibamen:
New bug report template with validation and Windows version


beeradmoore:
Good call, will check them after work.

The number of templates submitted with the placeholder text still 💀. Hopefully that helps

Saibamen:
You can see preview by clicking here:
image

beeradmoore:
Ah I didn't know about that, thanks! 

In that preview mode text input is disabled, I assume that's just the preview mode?

Saibamen:
Yes, readonly is for preview only.

You can check real example here: https://github.com/Saibamen/github-playground/issues

beeradmoore (approved):

beeradmoore:
Perfect, thanks for that

PR #393, opened 2025-02-16
Add Portable Configurations In sln And csproj

Zagrthos:
### Summary
Adds the Portable and Portable_Dev definitions right into the .sln and .csproj files to allow for better debugging.

### Detailed Changes
While looking through a fix for an issue only occurring in portable instances I found it hard to get a debug build for portable to work. So I created both of the new configurations and also adjusted the building package to use the new configuration.

Only the PORTABLE config is optimized, PORTABLE_DEV allows debugging.

### Notes
I also refactored the preprocessor directives in Storage.cs to be more simple to understand.

Tested for functionality.

### DIFF
--- a/package/build_Portable.cmd
+++ b/package/build_Portable.cmd
@@ -49,7 +49,7 @@ Set-Content -Path '%csproj_file%' -Value $content; ^
 dotnet publish "%csproj_file%" ^
 	--runtime win-x64 ^
     --self-contained ^
-    --config PORTABLE ^
+    --configuration PORTABLE ^
     -p:PublishDir=bin\publish\portable\ || goto :error
 
 
--- a/package/build_Portable.cmd
+++ b/package/build_Portable.cmd
@@ -49,7 +49,7 @@ Set-Content -Path '%csproj_file%' -Value $content; ^
 dotnet publish "%csproj_file%" ^
 	--runtime win-x64 ^
     --self-contained ^
-    --configuration PORTABLE ^
+    --configuration Release_Portable ^
     -p:PublishDir=bin\publish\portable\ || goto :error
 
 
--- a/package/build_Portable.cmd
+++ b/package/build_Portable.cmd
@@ -49,7 +49,7 @@ Set-Content -Path '%csproj_file%' -Value $content; ^
 dotnet publish "%csproj_file%" ^
 	--runtime win-x64 ^
     --self-contained ^
-    -p:DefineConstants="PORTABLE" ^
+    --config PORTABLE ^
     -p:PublishDir=bin\publish\portable\ || goto :error
 
 
--- a/src/DLSS Swapper.csproj
+++ b/src/DLSS Swapper.csproj
@@ -12,7 +12,7 @@
     <EnableMsixTooling>true</EnableMsixTooling>
     <WindowsPackageType>None</WindowsPackageType>
     <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
-    <Configurations>Debug;Release;Portable;Portable-Dev</Configurations>
+    <Configurations>Debug;Release;Portable;Portable_Dev</Configurations>
     <PackageIcon>Asssets\icon_256.png</PackageIcon>
     <ApplicationIcon>Assets\icon.ico</ApplicationIcon>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
--- a/src/DLSS Swapper.csproj
+++ b/src/DLSS Swapper.csproj
@@ -7,12 +7,12 @@
     <RootNamespace>DLSS_Swapper</RootNamespace>
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <Platforms>x64</Platforms>
-    <RuntimeIdentifiers>win-x64</RuntimeIdentifiers>
+    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
     <UseWinUI>true</UseWinUI>
     <EnableMsixTooling>true</EnableMsixTooling>
     <WindowsPackageType>None</WindowsPackageType>
     <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
-    <Configurations>Debug;Release</Configurations>
+    <Configurations>Debug;Release;Portable</Configurations>
     <PackageIcon>Asssets\icon_256.png</PackageIcon>
     <ApplicationIcon>Assets\icon.ico</ApplicationIcon>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
@@ -24,7 +24,7 @@
     <PublishReadyToRunShowWarnings>true</PublishReadyToRunShowWarnings>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(Configuration)'=='Release'">
+  <PropertyGroup Condition="'$(Configuration)'=='Release' OR '$(Configuration)'=='Portable'">
     <Optimize>True</Optimize>
   </PropertyGroup>
 
--- a/src/DLSS Swapper.csproj
+++ b/src/DLSS Swapper.csproj
@@ -12,7 +12,7 @@
     <EnableMsixTooling>true</EnableMsixTooling>
     <WindowsPackageType>None</WindowsPackageType>
     <WindowsAppSDKSelfContained>true</WindowsAppSDKSelfContained>
-    <Configurations>Debug;Release;Portable</Configurations>
+    <Configurations>Debug;Release;Portable;Portable-Dev</Configurations>
     <PackageIcon>Asssets\icon_256.png</PackageIcon>
     <ApplicationIcon>Assets\icon.ico</ApplicationIcon>
     <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
--- a/src/DLSS Swapper.csproj
+++ b/src/DLSS Swapper.csproj
@@ -29,11 +29,11 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
-    <DefineConstants>$(DefineConstants);Debug;Portable</DefineConstants>
+    <DefineConstants>$(DefineConstants);DEBUG;PORTABLE</DefineConstants>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
-    <DefineConstants>$(DefineConstants);Release;Portable</DefineConstants>
+    <DefineConstants>$(DefineConstants);RELEASE;PORTABLE</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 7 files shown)


beeradmoore:
@Zagrthos , I wonder if this can be done simpler? 

Add a `DLSS Swapper.Portable.sln`, which uses `DLSS Swapper.Portable.csproj` and that csproj will then set defined constants and then include the main csproj. I think all of that is possible.

That way we don't have to deal with debug/release/portable/portable_dev and can just keep debug/release.

The downside is you can't quick swap inside the IDE.

Does that make sense, or am I over thinking it?

Zagrthos:
@beeradmoore Well that would be another option. The main one I am used to is to just define them like here and then you can quickly switch over. Since no one actually reads the sln files I wouldn't see an issue doing it this way.

I am open to suggestions tho - just did the thing I'd do. Maybe I am missing something or an easier way to do this. 🙂

beeradmoore:
My main concern is with complexity. Having to remember to write `PORTABLE_DEV` every time I write `PORTABLE` can lead to issues if it is missed.

I don't think `Debug` configuration is used if we use `Portable_Dev` (like it does not inherit it). So now in all cases we want to use `Debug` or `Release` in the csproj (although currently not many) we have to consider the other instances of it.

That is the only reason I think that using `<Import Project="DLSS Swapper.csproj" />` could be useful to keep it as one config and two seperate solutions. That is assuming it works for this project type, I had only done it with old Xamarin.Forms projects for white labelled apps.

Another downside is the portable csproj has to be manually maintained.  If you are editing from the portable solution and you add a new image it will add it to the portable csproj and then won't exist in the installer version. 

There may be something that can do both though (pause while I get the screenshot from my Windows PC)

Zagrthos:
The Package / Unpackaged solution is mainly used for Microsoft Store Distribution projects (because when you run as Package it creates an msix package, unpackaged is self-contained). I dunno if you want to switch to the .msix file format instead of Inno Setup (could be easier to publish to MS Store probably?).

I would, personally, move away from double solutions and csproj files because of (you mentioned it) the double maintaining complexity which is (IMHO) higher than if you add
```cs
#if PORTABLE || PORTABLE_DEV
#endif
```

The only place I see where PORTABLE_DEV is really needed seems to be in the Storage.cs file. There you set the path of the settings manually to DEBUG when it's a debug build. Maybe we can find a solution to only solve this single place? With a bool or something based on the file name?

In another private project I differed the versions (RELEASE, DEBUG, TESTING) with different file-suffixes like DLSS-Swapper-DEV.exe and then checked if the current executable ends with one of them to set a bool (isDebug) which changed the behavior of the app.

beeradmoore:
> The Package / Unpackaged solution

I don't mean changing to packaged/unpackaged. I mean is there a mechanism we can do to make normal and portable show up in that dropdown?

I feel like long ago I had that but when I ripped out packaged apps I removed it as well. But I don't recall if I had it there, or if that is in my head.

> The only place I see where PORTABLE_DEV is really needed

The problem is when we do something new elsewhere and expect it to work but have forgotten to use `PORTABLE_DEV ` as well. It's why `#if DEBUG` and `#if !DEBUG` work great because it doesn't matter if it's portable or not, and then for something you are explicitly doing portable in you just use `#if PORTABLE`.

The `PORTABLE` was used a lot more when the app was a packaged app, but we moved away from that when the Windows Store started rejecting updates. 

> Maybe we can find a solution to only solve this single place? With a bool or something based on the file name?

I think the ideal solution is to have debug and release as the only configurations, but something else toggles portable or not portable. I just don't know if that solution is possible 😅

beeradmoore:
I had a look at the dropdown I was thinking of,

Screenshot 2025-02-23 093806

This is controlled by `launchSettings.json`. 

```json
{
  "profiles": {
    "DLSS Swapper": {
      "commandName": "Project"
    },
    "DLSS Swapper (Portable)": {
      "commandName": "Project"
    }
  }
}
```

But unfortuantly this is done post build so we can't change things in the cspoj like defined constants. I tried copilot to see what it says and it said I can set environment varaibles in launchSettings and then read them from the  csproj to then set defined constants. That does not seem to work so it is possible that it made it up 😂

I had a look at the original way you were doing things and I think the best way currently is to:

1. Go into config manager and add new config `Debug_Portable` and copy settings from `Debug`, do the same for `Release_Portable` with `Release.

2. We only have one csproj setting for release, but it can be updated to
```xml
  <PropertyGroup Condition="'$(Configuration)'=='Release' OR '$(Configuration)'=='Release_Portable'">
    <Optimize>True</Optimize>
  </PropertyGroup>
```

3. Add the following to csporj to ensure `DEBUG` defined constants exists for `Debug_Release` configuration. Same for release. This also adds `PORTABLE`

```xml

    <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
        <DefineConstants>$(DefineConstants);DEBUG;PORTABLE</DefineConstants>
    </PropertyGroup> 
    
    <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
        <DefineConstants>$(DefineConstants);RELEASE;PORTABLE</DefineConstants>
    </PropertyGroup>
```

4. Test it is working. I used the following config to jump between different ones and it gives us the option of `DEBUG`, `RELEASE`  which work for `Debug_Portable` and `Release_Portable` as well as the PORTABLE`. It also gives addtional properties of `DEBUG_PORTABLE` and `RELEASE_PORTABLE`. However usage of these should probably be avoided because 

```csharp
#if DEBUG_PORTABLE
// debug portable code
#else
// This is debug, release, and release portable which is likely not what we intended.
#endif
```

Test code to see it live update in VS is,
```csharp
#if DEBUG
    Debugger.Break();
#endif

#if RELEASE
    Debugger.Break();
#endif

#if DEBUG_PORTABLE
    Debugger.Break();
#endif

#if RELEASE_PORTABLE
    Debugger.Break();
#endif

#if PORTABLE
    Debugger.Break();
#else
    Debugger.Break();
#endif
```

Zagrthos:
I had a look too the last few days and... probably ended up the same resolution you described in the post above lol.

Yet I think we should *maybe* stay with `Portable` and `Portable_dev` instead of defining another constant named `Portable` (which basically is for what both of them were introduced already).

In Storage.cs we still have the issue of defining a specific `DEBUG` path which only should be active when `PORTABLE` and `DEBUG` are defined. With the proposed changes of defining a single `PORTABLE` symbol when using `DEBUG` or `RELEASE` however we can't differ anymore between the two symbols and the `PORTABLE` storage location can't be resolved in VS -> Resulting in a permanent error being visible. 🙂

beeradmoore:
I don't like `Portable` and `Portable_Dev` as it doesn't reflect the same thing as `Debug` and `Release` but for `Portable` mode apps. It does make sense that you use that mode for developing portable app, but it is more specifically debug mode targeting a portable configuration.

> In Storage.cs we still have the issue of defining a specific DEBUG path which only should be active when PORTABLE and DEBUG are defined. 

```csharp
#if DEBUG && PORTABLE
// debug portable
#elif DEBUG && !PORTABLE
// debug installer
#elif RELEASE AND PORTABLE
// release portable
#elif RELEASE && !PORTABLE
// release installer
#endif 
```

I am pretty sure`&& PORTABLE == false` also works which I'd prefer for visibility

But also there is this horrible to read nonsense 😂
```csharp
#if DEBUG
#PORTABLE
// debug portable
#else
// debug installer
#endif
#elif RELEASE
#PORTABLE
// release portable
#else
// release installer
#endif
#endif
```

But I'd take that first one any day of the week


Unless I misunderstood the issue you're having?

Zagrthos:
Okay I may said it wrong but I implemented the thing you suggested here

> ```xml
>     <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
>         <DefineConstants>$(DefineConstants);DEBUG;PORTABLE</DefineConstants>
>     </PropertyGroup> 
>     
>     <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
>         <DefineConstants>$(DefineConstants);RELEASE;PORTABLE</DefineConstants>
>     </PropertyGroup>
> ```

and pushed it to the branch - well turns out *somehow* the Storage.cs thing still doesn't work and apply when Portable is defined. So either I got an issue in my implementation or something else is wrong. 🤔

---

> I don't like `Portable` and `Portable_Dev` as it doesn't reflect the same thing as `Debug` and `Release` but for `Portable` mode apps. It does make sense that you use that mode for developing portable app, but it is more specifically debug mode targeting a portable configuration.

What I meant here was having them still defined *and* in use but no single PORTABLE definition because then it suddenly stops working (or I do have a big oversight of what actually happens here - can't get this to work).

beeradmoore:
I jumped into storage to have a look and I saw what you meant by it not working right. I think the defined constants are case sensitive.

```xml
  <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
    <DefineConstants>$(DefineConstants);Debug;Portable</DefineConstants>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
    <DefineConstants>$(DefineConstants);Release;Portable</DefineConstants>
  </PropertyGroup>
```

should be 

```xml
  <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
    <DefineConstants>$(DefineConstants);DEBUG;PORTABLE</DefineConstants>
  </PropertyGroup>

  <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
    <DefineConstants>$(DefineConstants);DEBUG;PORTABLE</DefineConstants>
  </PropertyGroup>
```

package/build_Portable.cmd
```diff
@@ -49,7 +49,7 @@ Set-Content -Path '%csproj_file%' -Value $content; ^
 dotnet publish "%csproj_file%" ^
 	--runtime win-x64 ^
     --self-contained ^
-    -p:DefineConstants="PORTABLE" ^
+    --configuration PORTABLE ^
```
beeradmoore, line 52: Because we have a new configuration, I believe this is meant to be `--configuration Portable_Release`

src/DLSS Swapper.csproj
```diff
@@ -24,11 +24,19 @@
...
     <Optimize>True</Optimize>
   </PropertyGroup>
 
-     <ItemGroup>
+  <PropertyGroup Condition="'$(Configuration)'=='Debug_Portable'">
+    <DefineConstants>$(DefineConstants);Debug;Portable</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='Release_Portable'">
+    <DefineConstants>$(DefineConstants);Release;Portable</DefineConstants>
```
beeradmoore, line 36: `Debug`, `Release` and `Portable` need to be capitalised inside `DefinedConstants` for them to work elsewhere.

Today I learned that defined constants is case sensitive, I just assumed it made them all caps anyway

beeradmoore (changes requested):

beeradmoore:
I also just found I can commit to a PR, so I'll fix those here for you and approve

beeradmoore (approved):

Zagrthos:
It all was because of the capitalization? Oh dear, .NET being .NET again.
Thanks for fixing and the merge then. Should get the debugging experience a little better now. 🙂

Issue #448, opened 2025-03-03
Traditional Chinese language support


beeradmoore:
> Hey there! This project has been a huge help to me! I’d love to share it with others who speak the same language as I do. But before that, the app needs to be translated, so I’d love to contribute by helping with the Traditional Chinese translation. I’m really looking forward to the project opening up for community contributions! 

 _Originally posted by @everland-3769 in [#425](https://github.com/beeradmoore/dlss-swapper/issues/425#issuecomment-2694626198)_

beeradmoore:
Closing this issue off, you will soon be tagged in #563 where there is more information about the future of translations.

Issue #425, opened 2025-02-23
[Feature request] Add translation support


beeradmoore:
### Describe what this adds or solves

DLSS Swapper is currently in English. There have been multiple requests to add support for other languages with members even offering to contribute to translations.

### Describe the solution you'd like

Setup the backing translation functionality and provide users a way to contribute translations without having to be fully fledged developers. This could come in the form of documnetation or external tools.

### Additional context

Current state of translations

DutraGames:
Hey @beeradmoore !

It looks like I wasn’t the only one who brought up the topic of multi-language support.  

Good luck with implementing the language system! I believe using a JSON file, for example, could be a viable solution. However, I’m not entirely familiar with the DLSS Swapper structure or how C# handles this functionality.  

Also, I really liked the revamped feature request format. It looks great—congratulations on the improvement!  

Once you manage to set up language support, please tag me. I’ll be ready to start translating into Brazilian Portuguese, as requested by our colleague in #261, as well as into Spanish.  

Thanks again, and keep up the great work!

beeradmoore:
I figured I needed a master issue to keep track of them all, and keep track of where I am in being able to have the core system working before others can do it.

DutraGames:
Hey @beeradmoore, how's it going?

Just a suggestion if it's feasible, instead of creating the folders outside of a parent folder, as shown in the image:

![Image for Folders](https://github.com/user-attachments/assets/b3204fe0-9a4d-4699-91c0-70e3c3513a3a)

Would it be possible to use WINUI 3 with the following layout:
![Image for lang](https://github.com/user-attachments/assets/db972351-0168-4e47-8333-57359fdfdd53)

beeradmoore:
Unfortunately not. Those folders are not created by DLSS Swapper but by one or a few of the nugets it loads

That is unless there is a setting or something I am unaware of to tell it where to load application dlls from 

I was excited to try the "build single executable" feature which spits out a single executable file with everything contained in it but it's a lie. It actually just extracts itself into local app data. Was doing it for portable apps but a portable app that leaves files behind almost defeats the purpose

beeradmoore:
But to add to this, translations will not live in those folders and they themselves will be their own folder. So at least seeing translation files will be easy!

DutraGames:
Thank you! I’ll be looking forward to contributing more and more to the project!

everland-3769:
[...]

Issue #410, opened 2025-02-20
[Feature request] File language 


odutra-dev:
**Is your feature request related to a problem? Please describe.**  
I apologize for opening an issue about this, but this is more of a curiosity than a problem.  

**Describe the solution you'd like**  
I have started exploring the DLSS Swapper source code, but I haven't been able to locate the file responsible for language settings. I'm particularly interested in the portable version, as I noticed it generates multiple language folders.  

**Additional context**  
I would like to commit to keeping the Portuguese and Spanish translations up to date, both in the README and within the software. Additionally, I'm willing to contribute to future updates whenever possible.

beeradmoore:
Hey @DutraGames , 
Those language folders are built automatically and not due to languages in the app.

The app is currently not setup for translations currently. Let me see on this weekend if I can setup the app to use multiple languages so then a translation can be done.

DutraGames:
Thank you for the clarification! I'll wait to see if adding support for multiple languages will be possible. If so, I'd be happy to contribute to the translation.  

Since my question has been answered, I will close this issue. Once again, I appreciate your attention!

DutraGames:
Hey @beeradmoore , I saw that you added this feature for the new version, so I will keep this issue open.  

I'll adjust the title to make it clearer for anyone reading.

beeradmoore:
Closing this issue off, you will soon be tagged in #563 where there is more information about the future of translations.

Issue #365, opened 2025-02-05
I want to add Turkish language to your app. Can I provide language support?


denizgokbas:
Your app is very nice and I would like to localize it to my language. Can I translate the text and menus in the app to my language?

beeradmoore:
Thanks for the request @denizgokbas , someone has requested Brazilian Portuguese as well.

When I can I'll change the app to support translations and then put out a request + guide for users to create translations in their own language.

H-Bole:
Looking forward to this feature!
期待此功能！

beeradmoore:
Closing this issue off, you will soon be tagged in #563 where there is more information about the future of translations.

PR #482, opened 2025-03-27
Implementing multilang system (Polish available)

Kur3k:
## Description of Changes

Extracted most of english strings into (Recommend to use resxmanager extension for VS for working with translation files)
Resoruces.resx
Translated to polish
Resoruces-pl.resx

On start selected language is CurrentCulture.
If not supported then english is default.
Language may be selected and saved in Settings section.
Settings support hot reload, which means you don't need to restart your application to apply changing your language.

obraz

[x] New feature

#410
#384 
#365 
#261 
#448 
#425


src/.editorconfig
Kur3k, line 1: Should've add this file to .gitignore

Kur3k:
[2025_03_27_21_01.xlsx](https://github.com/user-attachments/files/19493642/2025_03_27_21_01.xlsx)
Current translation status.

src/NetworkTesterWindowModel.cs
```diff
@@ -858,4 +859,42 @@ void CancelCurrentTest()
             _cancellationTokenSource.Cancel();
         }
     }
+
+    #region LanguageProperties
+    [LanguageProperty] public string NetworkTesterWindowText => ResourceHelper.GetString("ApplicationTitle") + " - " + ResourceHelper.GetString("NetworkTester");
```
Kur3k, line 864: Use string interpolation.

src/FailToLaunchWindowModel.cs
```diff
@@ -1,35 +1,61 @@
...
     [RelayCommand]
     void CopyText()
     {
         var package = new DataPackage();
         package.SetText(SystemData);
         Clipboard.SetContent(package);
     }
+
+    #region TranslationProperties
+    [LanguageProperty] public string ApplicationFailToLaunchWindowText => ResourceHelper.GetString("ApplicationTitle") + " - " + ResourceHelper.GetString("FailedToLaunch");
```
Kur3k, line 32: Use string interpolation.

src/.editorconfig
beeradmoore: `.editorconfig` should not be in `.gitignore`

beeradmoore:
Hey @Kur3k , thanks for this. Looks good, but lots of things for me to dig through. I'll have a closer look after v1.1.7 is released.

Just initial thought after scrolling the page as it tried to kill my browser, I saw some things were added to `DLLRecord`. They were generic words like "Delete", "Cancel", etc. Should we do a single class that is bindable that holds names of generic words that are not page/context specific?

Kur3k:
> Hey @Kur3k , thanks for this. Looks good, but lots of things for me to dig through. I'll have a closer look after v1.1.7 is released.
> 
> Just initial thought after scrolling the page as it tried to kill my browser, I saw some things were added to `DLLRecord`. They were generic words like "Delete", "Cancel", etc. Should we do a single class that is bindable that holds names of generic words that are not page/context specific?

Kur3k:
@beeradmoore 
Thought about it is sense of creating something like LocalizedViewModelBase that inherits Observable object and implements common operations.

beeradmoore:
I forgot I had a prototype of localisation. I am not in a place to build/test that it is still working as intended, I can try later. [WinUI3-LocalizeTest](https://github.com/beeradmoore/WinUI3-LocalizeTest) is where I was testing some stuff. The `.pri` files I was working on were not needed. This sample puts `.resw` files in `Strings/{language}/Resources.resw`.

`MainWindow.xaml` has some examples of loading strings. 

This would use the ResourceHelper to load the string `Test1/Text`
```xaml
<TextBlock Text="{x:Bind local:ResourceHelper.GetString('Test1/Text')}" />
```

This would load the string `Test1` and put its value in the `Text` property of this `TextBlock`
```xaml
<TextBlock x:Uid="Test1" />
```

Having values set like `Test1/Text` means we can also adjust other properties (width, height, image source, etc) from the resource file.

Unsure about your implementation (have not looked through it fully) but this can also fallback to default language when translation is missing.

I am unsure what happens if you change `Microsoft.Windows.Globalization.ApplicationLanguages.PrimaryLanguageOverrride` live, if it updates all the references.

Kur3k:
@beeradmoore
`<TextBlock Text="{x:Bind local:ResourceHelper.GetString('Delete')}" />`
Works the same, when translation is missing it's also using default english translation.
However:
`<TextBlock Text="{x:Bind local:ResourceHelper.GetString(''Delete')}" />`
Works per view initialization and won't change in time.
We can do it in two ways, like hardcode everything in code like this, but it will require application restart every time to change language.
Storing translation properties in ViewModels allow to reload translations in application live during runtime.

beeradmoore:
Ah true. I remember from when I [requested support](https://github.com/microsoft/WindowsAppSDK/issues/5178#issuecomment-2676665457) someone had it auto-loading, but looking back at it they were reloading the whole page.

Binding per view model should work ok. I don't have specific idea of how that should be handled. It could be nice to have a single immutable object per view model always set to its `Strings` or `Localisation` property which would be `LibraryPageLocalisation` for `LibraryPage` and then `LibraryPageModel`. The `Localisation` property would be observable and if we bind the UI like

```
<TextBlock Text="{x:Bind ViewModel.Localisation.Delete, Mode=OneWay}" />
```

Then in `LibraryPageModel` can have a weak messaging (have not used one before, unsure what you used above) thing setup to handle language changed event which would just re-create the `Localisation` property to be a new `LibraryPageLocalisation` object.

At least as far as I know updating `ViewModel.Localisation` then causes the entire `ViewModel.Localisation.Delete` value to be reloaded.

At least that way it keeps a whole bunch of noise out of the view model and keeps a whole bunch of the codegen less as now its just +1 thing per view model that is observable, not per string we display on the screen.

Unsure if this has downsides, I am happy to hear of alternatives.

beeradmoore:
In terms of getting users to work on translations I was going to build a seperate repo which exposes itself as a web interface for user to see current translations, write their own and then submit a new/updated `Resource.resw` in an issue or if they want as a PR.

Kur3k:
@beeradmoore
SharedTranslationPropertiesViewModel +
SpecificPageTranslationViewModel with properties per specific pages sounds great

> In terms of getting users to work on translations I was going to build a seperate repo which exposes itself as a web interface for user to see current translations, write their own and then submit a new/updated `Resource.resw` in an issue or if they want as a PR.

I think uploading excel file for live view in google docs with template and allow creating Sheets foreach language will be good enough.

beeradmoore:
> SharedTranslationPropertiesViewModel +
SpecificPageTranslationViewModel with properties per specific pages sounds great

Awesome

> I think uploading excel file with restrictions for live view in google docs with template and allow creating Sheets foreach language will be good enough.

Google Sheets could work, but we'd need a way where users can't edit other peoples input but can use it as a base (to prevent the form being defaced) but they can also export in the right format (or we have a tool that converts csv to resw which isn't too hard) so then it can be submitted.

Kur3k:
@beeradmoore
resXmanager has option to export translation files to excel and import them back
current template looks like this but it will be extending with extracted strings:
> [2025_03_27_21_01.xlsx](https://github.com/user-attachments/files/19493642/2025_03_27_21_01.xlsx) Current translation status.

Kur3k:
Need to optimize them a little bit
But i think just showing the template and allow users to send links in the issues to their translations may be good enough instead of creating special api.

beeradmoore:
So long as the barrier to entry for the translator is minimal as they are likely not developers. That is why I originally thought of a weblink where they can just see what is done, fill in the remaining, hit a button and then send that file back here without even needing to use git (or they can use git if they want contributor status).

That is a problem for later down the line though

beeradmoore:
@Kur3k, can you let me know when you are ready for this to be reviewed

beeradmoore:
I played around with that other library. The only thing I see it adding that we could benifit from is dynamically loading languages at runtime. Even that it looks like it is just loading the `*.resw` files and parsing them manually as xml files. Likely something that can be implemented here.

beeradmoore (approved):

Issue #317, opened 2025-01-30
[BUG] Import DLL doesn't work...


Rool-IO:
**Describe the bug**
Import DLLs, can't see them in the library nor swap them in...

**To Reproduce**
Import DLLs, look vainly for them, abandon hope

**Expected behavior**
For it to work

**Logs**
Doesn't contain anything relevant that I can see..

**Screenshots**
Nothing to see, that's kind of the point :/

**Additional context**
It used to work in 1.0.5, sad regression

MysteriousAeon:
I assume you're trying to import a DLSS 4 DLL?

essi2:
oh this is a known issue, well guess it is manually swapping for now then

beeradmoore:
Known issue reported in v1.1 notes. I should have kept it in all the release notes

beeradmoore:
WIP of this.

TL;DR; New functionality is bigger and better than last (at least I hope)

You can now select multiple files. Previously this was a single file. Additionally you can mix and match with zip and dll in that selection. You can't import a directory though. So if you have your DLLs in different sub-folders you will have to go into each folder and select all and import.

The definition of "downloaded" and "imported" is getting a little muddy. I am slowly adding notes [to the wiki](https://github.com/beeradmoore/dlss-swapper/wiki/Troubleshooting#imported-dll-looks-the-same-as-a-downloaded-dll) to explain this. Essentially "downloaded" means it "known" and "imported" means "unknown". 

It is weird if you import a DLL and it shows up the same as you downloaded it. I don't think the user will realise this though. 

Some of the other changes in the backing code is also leading towards an auto-import system. If you download a game with DLSS 310.3, then DLSS Swapper should just import that. It can be an imported/unknown DLL until our manifest updates at which point it becomes downloaded/known. Win-win.

When something gets imported a few things can happen:
- If the file is a zip, and is a known zip hash it just copy/pastes the file and marks it as downloaded. This should be a super fast import.
- If the file is a zip, and not a known zip hash we go through every dll and import them as though you had selected it from the disk
- If the file is a dll, and is a known dll hash it is zipped and marked as downloaded. In the code the `zip_md5_hash` is no longer relevant to the zip on the disk and is only relevant for downloading zip and verifiying its intergrity. I don't see this causing problems.

[...]

PR #484, opened 2025-03-29
Bugfix/import dlls

beeradmoore:
## Description of Changes

This fixes the broken DLL import system. It also adds a few improvements such as being able to import a zip containing multiple recrods, as well as being able to select multiple records at once from the file picker.

This change also overhauls how manifests are loaded and used to create the main `ObservableCollectionView`.

Failing to load manifests will now report an error and exit the app.

### Additional changes that should have been their own branches
- Disabled CA2201 in .editorconfig to not warn about using `Exception` class.
- Made the GitHub app update checker to still run on launch, but skip if it was ran within 30 minutes.
- Added loading indicator in library page much like games page.
- Changed library SelectorBar to be generated in code not in xaml. This paves the way for us to change the order of them in the future.
- Fixed padding at the bottom of the Library page 
- Fixed typos of "Github" to "GitHub"
- Improved wording aroud deleting a DLL record.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Issues Resolved

#317

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -1,6 +1,6 @@
 ### From: https://github.com/dotnet/aspnetcore/blob/main/.editorconfig
 ### Last sync: c323953
-### Disabled: CA2007, IDE0073, SYSLIB1054
+### Disabled: CA2007, IDE0073, SYSLIB1054, CA2201
 ; EditorConfig to support per-solution formatting.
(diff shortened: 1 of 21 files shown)


beeradmoore (commented):
Self re-reviewed

Issue #242, opened 2024-11-03
[BUG] Updating game libraries from settings page does not cause game library to reload


beeradmoore:
**Describe the bug**
Now that game page remains loaded after its initial load we don't have a way to re-load it where required.

**To Reproduce**
1. Going to settings
2. Disalbe Steam
3. Go back to games
4. Steam will still show

**Expected behavior**
Steam games would disapear. 

**Screenshots**
N/A

**Additional context**
We could potentially add a "needs reload" to the library which would cause a force reload on that library (or instantly hide that library) depeding on the before/after state.

Could also maybe handle it in some sort of on appearing state of the page.

PR #485, opened 2025-03-29
[Bugfix] Refreshing game library on settings changed.

Kur3k:
## Description of Changes

Changing settings reflects game library.

- [x] Bug fix

Building new list from cache shouldn't last long.
It's rebuilding the whole games library.

Another option is to compare settings enum after change with previous enum to get GameLibrary enum that was actually removed, and then pass it in event args to remove only elements from specified GameLibrary and get them back from the cache, but don't know if is it such a big deal.

## Issues Resolved
[#242](https://github.com/beeradmoore/dlss-swapper/issues/242)

### DIFF
--- a/src/Data/GameManager.cs
+++ b/src/Data/GameManager.cs
@@ -293,9 +293,12 @@ internal partial class GameManager : ObservableObject
     {
         lock (gameLock)
         {
+            // TODO: Cancel loading of games here
             _synchronisedAllGames.Clear();
 
-            App.CurrentApp.RunOnUIThread(_allGames.Clear);
+            App.CurrentApp.RunOnUIThread(() => {
+                _allGames.Clear();
+            });
         }
     }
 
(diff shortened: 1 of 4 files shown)


beeradmoore:
Any issue if I come in and change the `event` to using `WeakReferenceMessenger` from `CommunityToolkit.Mvvm.Messaging` namespace? I'd like to experiment with how messaging systems work and this seems like a good canidate.

Kur3k:
> Any issue if I come in and change the `event` to using `WeakReferenceMessenger` from `CommunityToolkit.Mvvm.Messaging` namespace? I'd like to experiment with how messaging systems work and this seems like a good canidate.

Won't change much in this case.
Both messenger (Settings) and the receiver (PageModel) are objects that are never disposed from memory during application runtime.
However WeakReferenceMessenger would perform better in the opposite scenario, cuz it handles unsubscribing from event automatically, but you can use it with no doubt.

beeradmoore:
> Won't change much in this case. 

Wasn't hunting perf or trying to solve other unexpected issues, just wanted to try something new in production app 😄

beeradmoore (approved):

PR #488, opened 2025-03-31
[Feature] Logical disk drives search.

### LINKED ISSUE #470: [Feature request] Add option to ignore all games from a specific drive
### Describe what this adds or solves

> So the drive I'd like to hide only has games that predate DLSS completely.

Via [this comment](https://github.com/beeradmoore/dlss-swapper/issues/281#issuecomment-2735240513)

### Describe the solution you'd like

Add an option in settings to ignore games from a specific drive. 

### Additional context

_No response_

Kur3k:
## Description of Changes
Rendering only games from specified in settings logical hard drives.

## Type of Change
- [x] New feature

## Issues Resolved

#470

### DIFF
--- a/src/Data/EpicGamesStore/EpicGamesStoreLibrary.cs
+++ b/src/Data/EpicGamesStore/EpicGamesStoreLibrary.cs
@@ -1,21 +1,16 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Management;
 using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using DLSS_Swapper.Helpers;
 using DLSS_Swapper.Interfaces;
-using Microsoft.UI.Xaml.Controls;
-using Windows.Foundation.Metadata;
-using Windows.Security.Authentication.OnlineId;
 
 namespace DLSS_Swapper.Data.EpicGamesStore
 {
-    internal class EpicGamesStoreLibrary : IGameLibrary
+    internal class EpicGamesStoreLibrary : GameLibraryBase<EpicGamesStoreGame>, IGameLibrary
     {
         public GameLibrary GameLibrary => GameLibrary.EpicGamesStore;
         public string Name => "Epic Games Store";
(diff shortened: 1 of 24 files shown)


src/Data/EpicGamesStore/EpicGamesStoreLibrary.cs
```diff
@@ -96,74 +57,53 @@ public async Task<List<Game>> ListGamesAsync(bool forceNeedsProcessing = false)
...
+            var manifestsDirectory = Path.Combine(epicRootDirectory, "EpicGamesLauncher", "Data", "Manifests");
+            if (Directory.Exists(manifestsDirectory) == false)
+            {
+                return games;
+            }
 
+            Dictionary<string, CacheItem> cacheItemsDictionary = await GetCachedDictionaryAsync(epicRootDirectory).ConfigureAwait(false);
 
-            var foundManifestFiles = Directory.GetFiles(manifestsDirectory, "*.item");
+            string[] foundManifestFiles = Directory.GetFiles(manifestsDirectory, "*.item");
```
beeradmoore, line 87: This project should be using var everywhere (as in multiple files, I don't want to have to tag them all), it seems every bit of code copied has reverted to the full class names. Are you able to revert these back to vars?

src/Pages/GameGridPage.xaml
```diff
@@ -172,6 +172,12 @@
 
     </Page.Resources>
     <Grid>
+        <i:Interaction.Behaviors>
+            <i:EventTriggerBehavior EventName="Loaded">
+                <i:InvokeCommandAction Command="{Binding LoadedCommand}" />
```
beeradmoore, line 177: What is the gain from adding a new dependency to remove a single page loaded event? If this was to be done it shouldn't be bundled in this PR as there is so much noise going on in here.

src/Data/GOG/GOGLibrary.cs
```diff
@@ -1,22 +1,19 @@
...
 using System.Linq;
-using System.Text;
 using System.Text.Json;
 using System.Threading.Tasks;
 using DLSS_Swapper.Helpers;
 using DLSS_Swapper.Interfaces;
 using Microsoft.Win32;
-using Serilog;
 using SQLite;
+using Windows.UI.StartScreen;
```
beeradmoore, line 12: What was this for?

beeradmoore (commented):
Thank you for the work, but there is so much noise and architecture changes that appear to be going on here that shouldn't be nessisary. 

Currently I don't plan on merging this PR. 

After playing wth the functionality I think a better approach would be to have a set of ignored paths than an enable/disable toggle per drive. I will have a play with that later today.

PR #508, opened 2025-05-01
Battle.net Library Support

### LINKED ISSUE #16: [Feature request] Battle.NET support
**Is your feature request related to a problem? Please describe.**
No Battle.NET games appear.

**Describe the solution you'd like**
Add support for scanning of games installed via Battle.NET

**Additional context**
Test games - COD Cold War

Shaydera:
#### Summary

This pull request addresses [Issue #16](https://github.com/beeradmoore/dlss-swapper/issues/16)  by introducing support for Battle.net game library within DLSS Swapper. It enables detection of installed Battle.net games, retrieves game metadata, and integrates them into the application's existing game management flow.

#### New Features

- **Battle.net Library Integration**  
  Implements `BattleNetLibrary`, which reads and interprets the Battle.net `product.db` to list installed and playable games.

- **Battle.net Game Representation**  
  Introduces `BattleNetGame`, a subclass of `Game`, including support for platform-specific ID and playability checks.

- **Game Title Resolution**  
  Adds `BattleNetGameTitles`, a static mapping that translates known internal Battle.net UIDs (e.g., `"fenris"`, `"wow"`) into user-friendly display names.

- **Protobuf Parsing Support**  
  Parses `product.db` via a compiled `.proto` definition to extract product metadata in a structured and robust manner.

#### Rationale for Protobuf Integration

Battle.net's `product.db` is a binary-encoded file using Google Protocol Buffers (Protobuf). To extract meaningful data - such as game UIDs, install paths, and product states - the application must parse this format with an appropriate schema.

- A `.proto` file was reverse-engineered to match the structure of `product.db`.
- This schema was compiled into `BattleNetProductDb.cs`, which is used to deserialize the file into structured C# objects.
- The `Google.Protobuf` NuGet package was added as a dependency to enable this runtime deserialization.

Using Protobuf ensures accurate and future-proof parsing of Battle.net's data format, as opposed to brittle binary parsing.

#### Package Changes

- Added:  
  ```xml
  <PackageReference Include="Google.Protobuf" Version="3.30.2" />
  ```

#### Behavior and Safeguards

- World of Warcraft variants and the Battle.net Agent are explicitly ignored due to compatibility issues or irrelevance.
- Games are only added if they are marked as installed and directory is found on disk.
- Game UID to title mapping uses a Directory-based fallback if no explicit title mapping is found.
- Existing cache entries are removed if a game is no longer detected.

#### Cover Image Support

Unlike existing integrations (e.g., Steam, Epic), the Battle.net implementation **does not support cover image loading** at this time.  

- **No documented local cache** of usable cover assets was identified.
- **Remote asset requests observed during runtime** (identified via packet sniffing) require authenticated session tokens, making them inaccessible for client-side fetching without user credentials.

As a result, Battle.net games currently fall back to a generic or blank cover and skip the `UpdateCacheImageAsync` step. This is by design and may be revisited if a public or unauthenticated method for asset retrieval becomes available.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Issues Resolved

#16


beeradmoore:
Oh, excellent!

I'll do more review on this over the weekend.

On game covers, is it safe to assume there is a very small number of games that are available here? Could we have a hard coded list which loads them from one of the other game library services?

Shaydera:
> On game covers, is it safe to assume there is a very small number of games that are available here? Could we have a hard coded list which loads them from one of the other game library services?

Yes, the number of available games is quite small — currently only 31 titles are listed on Battle.net, with 5 of them added in the last \~2 months. These newly added titles are not yet included in the existing game title mapping, as I couldn’t find any public source for their UIDs, and I don’t personally own them to check manually.

The titles I’ve added so far were extracted from [Lutris static mapping](https://github.com/lutris/lutris/blob/d0a8062e2096a15bef7015a7107b5b083b2551cb/lutris/services/battlenet.py#L32), which works similiar.

Because of this, using a static mapping to load covers from another service wouldn’t currently work for these newer games, for the same reason. 
Additionally, franchise titles for *StarCraft* and *Warcraft* are exclusive to Battle.net and aren't available on other platforms, though these games don’t support DLSS/FSR anyway, so it’s less of a concern.

For the remaining titles, using a hardcoded mapping to another source as a fallback does seem like a reasonable workaround.
Steam would be the best choice in this case, as the titles that are available on other platforms are always available on Steam.

src/Data/BattleNet/Proto/battle_net_product_db.proto
beeradmoore, line 1: Is the info used to build the proto file created by you, or from another source?
If from another source can we leave a comment in the file so we can remember where to update it from if something changes?

src/Data/BattleNet/BattleNetGameTitles.cs
```diff
@@ -0,0 +1,54 @@
+using System.Collections.Generic;
+using System.IO;
+using DLSS_Swapper.Data.BattleNet.Proto;
+
+namespace DLSS_Swapper.Data.BattleNet;
+
+internal static class BattleNetGameTitles
+{
+    private static readonly Dictionary<string, string> _uidTitles = new()
```
beeradmoore, line 9: This is where I was thinking of having the map of thumbnail images from either [steamgriddb.com](https://steamgriddb.com) or [igdb.com](https://igdb.com). I don't mind manually setting up the list and maintaining it going forwards.

src/Data/BattleNet/BattleNetGameTitles.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            return title;
+        }
+
+        if (string.IsNullOrEmpty(product.Settings.InstallPath))
+        {
+            return product.Uid;
+        }
+
+        var dir = new DirectoryInfo(product.Settings.InstallPath);
```
beeradmoore, line 51: I think we should put an error log here, something like `Battle.Net game title not found for uuid {x} in install path {y}`. That way if people complain about something in the future the relevant data should be logged for us to get and update.

src/Data/BattleNet/BattleNetLibrary.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        var games = new List<Game>();
+        var tempFile = Path.GetTempFileName();
+
+        try
+        {
+            File.Copy(_productDbPath, tempFile, true);
+            await using var fs = new FileStream(tempFile, FileMode.Open, FileAccess.Read, FileShare.Read, 4096,
+                FileOptions.DeleteOnClose);
```
beeradmoore, line 53: I didn't know about `DeleteOnClose`. Thanks for sharing!

beeradmoore (commented):
Thanks for this PR. Fixes a very long standing issue.

I'll do a few formatting changes before I merge to main. 

As for the requested changes, I am more than happy to do them, just let me know. I am going to release v1.1.7 tomorrow. If this doesn't make the cut I'll make it v1.1.8 for release during this week to do before v1.2 (DLSS presets release).

This PR did show two other problems that I had not thought of.
1. Because of how enabled libraries are stored, Battle.Net will be disabled by default for existing users.
2. I don't like how the games library list is as I prefer manually added games at the bottom. But then the rest of the list is sort of random. In another branch I'll look at adding the ability to sort them.

src/Data/BattleNet/BattleNetGameTitles.cs
Shaydera: Great idea! Done.

src/Data/BattleNet/Proto/battle_net_product_db.proto
Shaydera: I derived it from the byte string at https://github.com/lutris/lutris/blob/d0a8062e2096a15bef7015a7107b5b083b2551cb/lutris/util/battlenet/product_db_pb2.py#L24

Ive added that information to the file.

src/Data/BattleNet/BattleNetLibrary.cs
```diff
@@ -0,0 +1,180 @@
...
+    private BattleNetLibrary()
+    {
+        var allUsersProfile = Environment.GetEnvironmentVariable("ALLUSERSPROFILE");
+        allUsersProfile ??= Environment.ExpandEnvironmentVariables("%ProgramData%");
+        _productDbPath = Path.Combine(allUsersProfile, "Battle.net", "Agent", "product.db");
+    }
+
+    public bool IsInstalled()
+    {
+        var agentPath = Directory.GetParent(_productDbPath)?.FullName;
```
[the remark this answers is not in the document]
Shaydera: Thanks for pointing that out, I would have never guessed that the agent persists after uninstall...... and thus did care to test.

src/Data/BattleNet/BattleNetGameTitles.cs
Shaydera: I’m happy to leave the implementation up to you, if that’s alright, since it seems you already have a clear idea of how you want to implement it and manage the files.

src/Data/BattleNet/BattleNetLibrary.cs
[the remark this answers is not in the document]
beeradmoore: Neither. I stumbled upon it accidentally when seeing if there were any other per-game registry changes we could use for some info. Uninstalled Battle.Net at some point and realised DLSS Swapper was still reporting it as installed.

src/Data/BattleNet/BattleNetGameTitles.cs
beeradmoore: No worries. I'll get to that tomorrow and get it in for v1.1.7

beeradmoore:
Thanks for those changes. I'll get that all into main tomorrow and will be part of the v1.1.7 release.

beeradmoore (approved):

beeradmoore:
Thanks for the work on this @Shaydera
Was a nice little bonus to include in the v1.1.7 release

beeradmoore:
@Shaydera , earlier when I had you change BattleNet detection to check for the intalled app, I have gone and reverted that part to what you originally had. If you uninstall the battlenet client the games remain installed so they should still be detectable.

PR #513, opened 2025-05-04
Feature/reorder libraries

beeradmoore:
## Description of Changes

Adds the ability to re-order game libraries. This also changes the saved structure of how enabled/disabled game libraries are stored. This fixes an issue where adding a new library would be disabled by default. Unfortuantly because this is the v1.1.7 when Battle.Net is also added it won't be enabled by default.

https://github.com/user-attachments/assets/bfb3fc75-068d-4a0b-8ced-6446f8e4cd90

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update


beeradmoore (commented):
.

PR #536, opened 2025-05-21
Update README.md

Tythesly:
## Description of Changes

Added Battle.net to supported game library section

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Documentation update


beeradmoore (approved):
Thanks @Tythesly , good catch!
I did a minor tweak of changing order so manually added is below battlenet, and then copied that changed to the other translated readmes

PR #564, opened 2025-06-22
Update Simplified Chinese translation

Rakau:
Changelog: Completed untranslated text and corrected some translations

Next, I will perform more refined localization of the translated text


beeradmoore:
Thanks for these updates!

PR #569, opened 2025-06-24
Feature/translations setup

odutra-dev:
## Description of Changes

This PR introduces two key enhancements to the **DLSS Swapper Translator Tool**:

1. **Auto Save Functionality**  
   - Adds automatic saving of user translations using `localStorage`.  
   - Translations are preserved between page reloads without requiring manual saving.  
   - Auto Save is enabled by default and can be toggled via a switch in the UI.

2. **Portuguese Language Support**  
   - Adds base support for translating the tool into Portuguese (`pt-BR`).  
   - Lays the groundwork for future localization and language selection.

These changes improve usability for translators and make the tool more accessible to Portuguese-speaking users.

---

## Type of Change

- [x] New feature

### DIFF
--- a/docs/translator/index.html
+++ b/docs/translator/index.html
@@ -1,41 +1,70 @@
 <!DOCTYPE html>
 <html lang="en" data-bs-theme="auto">
-    <head>
-        <meta charset="UTF-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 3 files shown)


beeradmoore (approved):

beeradmoore:
Thanks! This is now in. I put the language selector name as "Português BR" because that is what is used on the readme, let me know if that should be somethng else to make more sense to people looking for it.

PR #584, opened 2025-07-06
Prevented import/export/add game from working when running as admin

beeradmoore:
## Description of Changes

This does not fix the issue of being able to add game, import/export dlls when running as admin. However it will error instead of crash which will reduce the support required from my end so I can focus on other things.

Folder 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

### DIFF
--- a/src/Pages/GameGridPageModel.cs
+++ b/src/Pages/GameGridPageModel.cs
@@ -107,6 +107,19 @@ public partial class GameGridPageModel : ObservableObject
     [RelayCommand]
     async Task AddManualGameButtonAsync()
     {
+        if (Environment.IsPrivilegedProcess)
+        {
+            var errorDialog = new EasyContentDialog(gameGridPage.XamlRoot)
+            {
+                Title = ResourceHelper.GetString("General_Error"),
+                CloseButtonText = ResourceHelper.GetString("General_Okay"),
+                DefaultButton = ContentDialogButton.Close,
+                Content = ResourceHelper.GetString("General_FeatureNotSupportedWhenAdmin"),
+            };
+            await errorDialog.ShowAsync();
+            return;
+        }
+
         if (Settings.Instance.DontShowManuallyAddingGamesNotice == false)
         {
             var dontShowAgainCheckbox = new CheckBox()
@@ -169,7 +182,6 @@ public partial class GameGridPageModel : ObservableObject
 
     async Task AddGameManually()
     {
-
         TextBlockBuilder textBlockBuilder = new TextBlockBuilder(ResourceHelper.GetString("GamesPage_ManuallyAdding_InfoHtml"));
 
         if (Settings.Instance.HasShownAddGameFolderMessage == false)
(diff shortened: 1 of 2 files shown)


Bankaix2:
Friend... in v1.2.0.2 despite giving privileges as "Adm" and running as "Adm" it does not allow adding games nor interacting with the translate editor.

beeradmoore:
This is correct. When the game is running as admin (not an administrator account) many features are disabled. 

Translation is disabled because when you want to save your translation it prompts a save file dialog box. When running as admin this would cause a crash. Rather than have people lose all their progress on translations you just can't launch the translation editor when running as admin.

Unless I am misunderstanding something, this is all by design to prevent crashes.

KAPTOHKA:
@beeradmoore
How can I run `DLSS-Swapper` without administrator rights so that a `yellow warning` does not appear, on a system without UTC, without antivirus, and with the only `administrator account`?
In it, all applications are launched from the `admin, by default`...
Maybe there is a `startup key` for this?

Shadetorrent:
Im sorry but this "fix" made the program unusable for me i cant add games now cause i'm using an admin account only

beeradmoore:
This has been reverted and properly fixed for the next version. 

Can you explain how everything runs as admin? My PC only has one account, the administrator account, but not all of my programs run in an elevated mode.

Trying to figure out why some peoples PCs do and some peoples PCs don't.

beeradmoore:
@Shadetorrent , here is the full post as to why this fix happened in the first place.

https://github.com/beeradmoore/dlss-swapper/issues/614

Shadetorrent:
> This has been reverted and properly fixed for the next version.
> 
> Can you explain how everything runs as admin? My PC only has one account, the administrator account, but not all of my programs run in an elevated mode.
> 
> Trying to figure out why some peoples PCs do and some peoples PCs don't.

I dont have a good answer to ur question idk, i always thought that if ur on the admin account then everything will run as admin as default, unless im not remembering something.

Anyways, i was able to get around it yesterday by reverting to older version, adding the game, then updating to latest version again; Thanks for all the work.

beeradmoore:
Glad you got a workaround for now.

AFAIK admin account is more that it allows you to is talk things. Standard user accounts cant. 

If you were to open task manager and go to the details tab there is a column called elevated (if there isn't you can right mouse click on table headers and enable it). DLSS Swapper had issues when running elevated. We called it "running as admin" in app because I thought the only way people got there is by right mouse clicking on the shortcut and going "run as admin".

That's why I am confused why so many systems are just doing this out of the box without people doing that

ShikieikiC:
@beeradmoore 

Hi, my friend. I also encountered the issue where the program doesn't function properly when run in Administrator mode and cannot add game paths. I noticed that you've already fixed this in a later version, which is great.

I’d like to add some context on why so many people tend to run this program in Administrator mode. It's not specifically about dlss-swapper—on some systems where security policies are more relaxed (like my personal desktop), I often disable UAC prompts and modify settings in the Local Group Policy Editor, specifically by disabling User Account Control: Run all administrators in Admin Approval Mode. This is actually a common practice for users who want to avoid system prompts during sensitive operations.So on such devices, all applications are effectively run in Administrator mode by default.

beeradmoore:
Thank you for the extra info

PR #592, opened 2025-07-13
Update Simplified Chinese translation

juij-fun:
Update Simplified Chinese translation (to align with my previous personal translation version)
Standardize some formatting of the original translation

### DIFF
--- a/src/Translations/zh-CN/Resources.resw
+++ b/src/Translations/zh-CN/Resources.resw
@@ -121,37 +121,37 @@
     <value>许可证 &amp; 致谢</value>
   </data>
   <data name="ApplicationTitle" xml:space="preserve">
-    <value>DLSS切换器</value>
+    <value>DLSS Swapper - DLSS 管理程序</value>
   </data>
   <data name="DiagnosticsPage_ClickToCopyDetails" xml:space="preserve">
     <value>点击复制以下详细信息</value>
   </data>
   <data name="DiagnosticsPage_WindowTitle" xml:space="preserve">
-    <value>诊断系统/系统信息</value>
+    <value>诊断信息</value>
   </data>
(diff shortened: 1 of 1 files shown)


beeradmoore:
There is some problems with the translation. I think it is some of the characters used for like `<i>` and stuff. If that is the issue I can resolve them. 

Out of curiosuty, did you do this in with the translation toolbox, or manually updated resw, or is this with web editor?

juij-fun:
> There is some problems with the translation. I think it is some of the characters used for like `<i>` and stuff. If that is the issue I can resolve them.
> 
> Out of curiosuty, did you do this in with the translation toolbox, or manually updated resw, or is this with web editor?

I edited the resw file. Since I couldn't enter the settings before, it has not been tested. If you can solve it directly, it is OK.

juij-fun:
> There is some problems with the translation. I think it is some of the characters used for like `<i>` and stuff. If that is the issue I can resolve them.
> 
> Out of curiosuty, did you do this in with the translation toolbox, or manually updated resw, or is this with web editor?

Re-edit and delete the wrong character <i>

beeradmoore:
Found the problem. Just a few broken XML tags. You can see the change [here](https://github.com/beeradmoore/dlss-swapper/pull/592/commits/af66e24870e4d525674309e58b1bc03968bd5971).

beeradmoore:
Thanks for the updates @juij-fun

juij-fun:
> Found the problem. Just a few broken XML tags. You can see the change [here](https://github.com/beeradmoore/dlss-swapper/pull/592/commits/af66e24870e4d525674309e58b1bc03968bd5971).

Thank you very much for your inspection and correction. I was really negligent before...

beeradmoore:
That is ok, it happens to us all :)

PR #608, opened 2025-07-14
Add Spanish translation

Deci8BelioS:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

### DIFF
--- a/src/Helpers/LanguageManager.cs
+++ b/src/Helpers/LanguageManager.cs
@@ -46,6 +46,7 @@ public class LanguageManager
             "en-AU",
             "en-GB",
             "en-US",
+            "es-ES",
             "pl-PL",
             "pt-BR",
             "ru-RU",
@@ -67,6 +68,7 @@ public class LanguageManager
             "en-AU" => "English (Australia)",
             "en-GB" => "English (United Kingdom)",
             "en-US" => "English (United States)",
+            "es-ES" => "Español", // Spanish
             "pl-PL" => "Polish",
             "pt-BR" => "Português BR",
             "ru-RU" => "Русский", // Russian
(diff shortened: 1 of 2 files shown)


beeradmoore:
Thanks, I'll try merge this in this weekend.

beeradmoore:
Hey @Deci8BelioS , this is missing one bit from the [translation guide](https://github.com/beeradmoore/dlss-swapper/wiki/Translation-Guide#creating-a-pull-request-preferred).

`src\Helpers\LanguageManager.cs` has two methods (`GetKnownLanguages` and `GetLanguageName`) which need to be updated to make the language availalbe in the app. I have added them, but if you could confirm these two questions before I merge it in.

Language code: I assume this is `es-ES`
Language name: this should be in the native language and is what is shown in the dropdown when the user picks a language, would this be `Español`?

Deci8BelioS:
> Hey @Deci8BelioS , this is missing one bit from the [translation guide](https://github.com/beeradmoore/dlss-swapper/wiki/Translation-Guide#creating-a-pull-request-preferred).
> 
> `src\Helpers\LanguageManager.cs` has two methods ( and ) which need to be updated to make the language availalbe in the app. I have added them, but if you could confirm these two questions before I merge it in.`GetKnownLanguages``GetLanguageName`
> 
> Language code: I assume this is Language name: this should be in the native language and is what is shown in the dropdown when the user picks a language, would this be ?`es-ES``Español`

Sorry, I forgot that part.

If it is correct it is `es-ES``Español`

beeradmoore:
No worries. Thanks for the work on this @Deci8BelioS . This will be in v1.2.1, whenever that happens to be.

PR #647, opened 2025-07-26
Fixed readme catalan translation

Dunocat639YT:
## Description of Changes

Fixed readme catalan translation for a better consistency.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Documentation update
- [X] Translation update

### DIFF
--- a/readmes/readme_ca.md
+++ b/readmes/readme_ca.md
@@ -53,7 +53,7 @@
 - [Ubisoft Connect](https://www.ubisoft.com/)
 - [Xbox App](https://www.xbox.com/)
 - [Battle.net](https://shop.battle.net/)
-- Afegit manualment mitjançant el botó `Afegeix partida`.
+- Afegit manualment mitjançant el botó `Afegeix joc`.
 
 ## Per què voldries canviar les DLL de DLSS al teu joc?
 
@@ -65,9 +65,9 @@ Aquesta eina **NO** et permet afegir DLSS a jocs que no el suporten.
 
 Aquesta eina **NO** garanteix que l'intercanvi del DLL del DLSS:
 
-- Millorar el rendiment del DLSS.
-- Reduir els artefactes del DLSS.
-- Oferir una experiència estable i sense errors.
+- Millori el rendiment del DLSS.
+- Redueixi els artefactes del DLSS.
+- Ofereixi una experiència estable i sense errors.
 
 En molts casos, podeu solucionar alguns problemes, en altres casos podeu impedir que s'iniciï un joc (fins que restaureu la DLL original, proporcionada a l'eina).
 


beeradmoore (approved):

beeradmoore:
Thanks again @Dunocat639YT

Dunocat639:
Thanks for merging it! These were my first pull requests ever, so I'm happy even though they were small.
I have a question: do you recommend using the portable or the installer version, and why?

beeradmoore:
I appreciate all additions large and small. It doesn't matter how good of a developer I am I'd never be able to do an accurate translation of the app/readme. So for that I am thankful. I also like that people can contribute to make the project better even if they are not working on features/bugfixes., etc.

Your PRs have all been top notch, I would not have known otherwise.

Personally I recommend installer version. Portable version is there for people who know what they are doing, and unfortuantly a lot of people don't know what they are doing.

Installer version is good, it's just there installed in your system. Updating is just running the installer again and the app is good to go. Plus launching it (at least without you doing anything) is as easy as hitting the windows button and typing "d l s s <enter>".

Portable version is intended for those who cant/dont want to install another appliation, or for those who want a portable version (eg. you can put it on a USB and just run it on your PC, friends PC, test bench, etc, without having to do much extra. But it comes at a drawback. There is no clean update. You have to download the new version, extract it, find the `StoredData` folder in the old version and put it in the new version.

If you want it in your startmenu or make a shortcut you need to add that yourself, and then when you update if the new version is in a new path you have to update all those shortcuts again.

With people not knowing what they are doing they will just move all the new files and place over the top of the old files. This can and does break things. It also makes it hard to diagnose issues when people say "I updated the portable version" but don't make it clear what that means. The reason this copy/paste system breaks is imagine the old app has a file called `some_module.44.dll` and the new app has `some_module.45.dll`. When you copy/paste both exist. When the app launches it does not selectively load the DLL that it was built for, it loads them all in the current directory. Now we have odd nonsense of loading both.

It can go the opposite way as well. What if the old app had `some_module.44.dll` but I have updated the program to not require `some_module` anymore. So now when you copy the files that file will still exist and be loaded. But the application is built to not want to use `some_module` anymore, and possibly just loading it is enough to cause problems.

Dunocat639:
Thanks for the response. I have another question: is there any specific reason to use DLSS Swapper over NVIDIA’s built-in DLSS override? Just curious if there are any benefits or differences between the two.

beeradmoore:
For DLSS presets, both should be the same. It's possible DLSS Swapper will override preset in games that are not whitelisted by NVIDIA, but I need to test that. Also sometimes it just doesn't work on some PCs 🤷‍♂️

For DLSS versions there are a few differences.

NVIDIA swaps at the driver level when the file gets loaded, DLSS Swapper swaps the actual file on the disk. 

NVIDIA can only swap what they have in the driver. DLSS Swapper only has what we have been able to source from SDKS/Games. If you don't update your driver as often as you should you won't get the latest. Additionally we had DLSS v310 (DLSS 4) and DLSS v310.3 before the driver did as game release/SDK release happened before driver update. That said a driver could update and have DLSS v310.4 before us.

NVIDIA DLSS override might be whitelisted (I need to test), we let you swap things even if it means the end result is a non-working game (its why we take a backup of the file first). With our approach you may be flagged for anti-cheat as you have altered files on the disk.


Additionally ours has had people not be able to enable DLSS after it is swapped. I guess some games do a hard check of "we shipped with these versions, this is not this version, do not load". NVIDIA override seems to work in these instances.

Minor thing, but we let you swap things that don't have a NVIDIA driver profile (kinda like whitelist but not the same). We let you point to a folder and if the file exists we let you swap it.   For example I can't swap DLSS in Tony Hawk 3+4 with NVIDIA app because it is not detected. I can with DLSS Swapper.

IMO at the end of the day they both have pros and cons. It can vary depending on the game, the system, the user, etc.

PR #650, opened 2025-07-28
Simplify language names in LanguageManager

Dunocat639YT:
## Description of Changes

I removed country-specific qualifiers from some language names (like German, French, and Italian) to make the language list more consistent.

Some languages (like Catalan, Spanish, Polish, or Russian) were already displayed without region tags, while others (like "German (Germany)") had them, even though only one version exists in the app. This inconsistency looked a bit odd and unnecessary.

I kept region tags only for languages that have clearly different regional variants that are used in the app (like Portuguese from Brazil, or English from the US, UK, and Australia). These need to be distinguished, but for the rest, just using the language name is cleaner and more consistent.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Translation update

### DIFF
--- a/src/Helpers/LanguageManager.cs
+++ b/src/Helpers/LanguageManager.cs
@@ -72,13 +72,13 @@ public class LanguageManager
             "ar-SA" => "اللغة العربية (المملكة العربية السعودية)", // Arabic (Saudi Arabia)
             "ar-SY" => " (سوريا) العربية", // Arabic (Syria)
             "ca-ES" => "Català", // Catalan
-            "de-DE" => "Deutsch (Deutschland)", // German (Germany)
+            "de-DE" => "Deutsch", // German (Germany)
             "en-AU" => "English (Australia)",
             "en-GB" => "English (United Kingdom)",
             "en-US" => "English (United States)",
             "es-ES" => "Español", // Spanish
-            "fr-FR" => "Français (France)", // French (France)
-            "it-IT" => "Italiano (Italia)", // Italian (Italy)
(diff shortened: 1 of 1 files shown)


Zagrthos:
I'd let them in honestly. Because how the language system works now it doesn't detect if, for example, someone uses `de-AT` (German with Austria dialect). Then it defaults to `en` because there's no single `de` locale and no matching `de-AT` one. This also makes it easier for people to recognize in which dialect one language is provided w/o consulting [search engine of your choice] first.

beeradmoore:
I had left it in because it is what the people submitting languages had done when asked how they want it to be displayed. I only speak one language so I am not sure on what the standard is.

Zagrthos:
Maybe create an Issue about this and let people vote by their voices which one is better?

beeradmoore:
I'll do a reddit and a twitter poll and see what people say.

beeradmoore (approved):
I have thought about this and I think I now agree. It should only exist if it needs it.

If there was no German language added but someoned added "de-CH" I would expect it to be listed as "German (Switzerland)" not "German". However if then someone added "de-DE" I would expect it to be listed as "German (Germany)".

If "de-DE" existed I think it is ok to be "German", but then when "de-CH" is added it then "German" would have to become "German (Germany)"

Dunocat639:
Should we do the same for the Portuguese language?

beeradmoore:
The Portuguese is "Português BR" not "Português". It could possibly become "Português (Brazil)" (or whatever this should be in native language) to be in line with other languages.

Issue #708, opened 2025-08-23
[BUG] "DLSS Preset" string not translated


Dunocat639:
### Describe the bug

Similar to issue #700, the string "DLSS Preset" string isn't translated (as seen in the imatge below), even after the #707 PR.

### Steps To Reproduce

1. Open DLSS Swapper
2. Click on a game
3. See not translated string

### Expected behavior

Be able to translate it thought the translation toolbox.

### Screenshots

<img width="750" height="682" alt="Image" src="https://github.com/user-attachments/assets/3291b685-ed95-4451-8671-9e5191ed0a0c" />

### DLSS Swapper version

v1.2.1.1

### Windows version

Microsoft Windows 11 Pro Version 24H2 (OS Build 26100.4946)

beeradmoore:
Those titles were never intended to be translated. Should they be? Like "DLSS" would stay the same but would it make sense to change the end of "DLSS Frame Generation".

I had always thought of it as a product name and wasn't to be translated, but I also am learning what other people need as I go 😅

Dunocat639:
These product names are usually translated. For exemple, in the oficial [NVIDIA website in Spanish](https://www.nvidia.com/es-la/geforce/news/dlss4-multi-frame-generation-ai-innovations/), they translate "Frame Generation" to "Generación de Fotogramas". Also they do it in the NVIDIA App, translating "Smooth Motion" to "Movimiento fluido".

So, for exemple in Spanish, it would be something like:
- DLSS Frame Generation -> Generación de Fotogramas de DLSS
- DLSS Ray Reconstruction -> Reconstrucción de Rayos de DLSS
- DLSS Preset -> Perfil de DLSS

And the same for FSR and XeSS.

beeradmoore:
Excellent. Thanks. In about an hour I'll be at my PC and I'll get this one through. This will also change the language on the library page for the tabs at the top as they are all fed from one place.

beeradmoore:
This is now resolved in #710 and is on main branch. All those texts ("DLSS", "DLSS Preset", "DLSS Frame Generation", "FSR 3.1 (DirectX 12)", "XeSS Frame Generation", etc) can now all be translated. This translation should now be used throughout the app such as tabs on library page, title of the popup when clicking library page, and upcoming filter update where you can filter to only show games with specific tech instead of "no swappable items"

PR #709, opened 2025-08-23
Update Catalan translations

Dunocat639YT:
## Description of Changes

Added new DLSS preset and general status strings, updated terminology from 'preset' to 'perfil' for DLSS.
Related to #707.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Translation update

### DIFF
--- a/src/Translations/ca-ES/Resources.resw
+++ b/src/Translations/ca-ES/Resources.resw
@@ -165,6 +165,15 @@
   <data name="DllZipInvalidNoDllFound" xml:space="preserve">
     <value>El fitxer zip de la DLL no era vàlid (no s'ha trobat cap DLL).</value>
   </data>
+  <data name="DLSS_Preset_AlwaysUseLatest" xml:space="preserve">
+    <value>Utilitzeu sempre l'última versió</value>
+  </data>
+  <data name="DLSS_Preset_Default" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


beeradmoore:
When you say updated the terminology for preset, you mean translated the word preset?

"DLSS preset" is like a product name. So long as people know what DLSS presets are in English but when they see it in another language they don't think it's something else then it's all good

Dunocat639:
Yes, I mean that I translated "DLSS preset". Even thought it's a product name, they usually translate it to other languages, as I explained in #708.
I also translated other missing translations.

beeradmoore (approved):

beeradmoore:
Thanks for yet another translation update @Dunocat639YT!

I also ran the powershell script in `extras\SortTranslations\sort.ps1`. It loads the translation file as xml and sorts alphabetically on the keys. This helps keep translaitons in order but also prevents some PRs where it looks like a bunch of things were removed and re-added because they were moved. 

I am unsure if you can run the script on your machine because of how powershell permissions work, but for me I open a powershell terminal at that window and write `pwsh .\sort.ps1` and look at the stuff that has changed. If its whitespace only I discard changes. If it has fixed the order of some other things I generally will commit those with a note that they were re-sorted.

PR #745, opened 2025-09-22
Update Simplified Chinese Translations (zh-CN)

EmiyaKatuz:
## Description of Changes

Update simplified Chinese translations. Entries that lack translations have now been translated. Moreover, translations for some specific entries have been updated to avoid grammatical ambiguities.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Translation update

PR #748, opened 2025-09-29

juij-fun:


beeradmoore (approved):

beeradmoore:
Hey @juij-fun , sorry for the delay. There was some conflicts from the #745 pull request that was merged just before it. I think I merged it accuratly (as accurate as I could with Google Translate 👀)

If things can be improved still I am always happy to review another PR.

PR #793, opened 2025-12-23
Improved persistence of main window position/size/state

andyeder:
## Description of Changes

If the application is maximized and then closed (whilst still in a maximized state), the application will then start maximized the next time it is launched. However, if you then attempt to restore ("un-maximize"), it has no prior size/position to restore back to and remains at an increased size, but not in a maximized state. It can make it awkward for the user to resize the window manually to the desired size.

This change adds main window tracking to ensure position/size/state is persisted correctly between launches of the application. Size and position is tracked with additional logic in WindowPositionRect to help streamline the process.

## Type of Change

- [x] Bug fix

## Issues Resolved

There is no related bug/issue listed in the repo.

### DIFF
--- a/src/MainWindow.xaml.cs
+++ b/src/MainWindow.xaml.cs
@@ -27,7 +27,7 @@ namespace DLSS_Swapper
 
         IntPtr _windowIcon;
 
-        private readonly WindowPositionRect _trackedWindow = new();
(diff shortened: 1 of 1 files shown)


beeradmoore:
Thanks for that @andyeder. Good catch. I added a few formatting changes (which seeing some of the old files I see I want to go fix my formatter in general)

PR #801, opened 2025-12-26
Add translation tools and complete missing strings across languages

peter9811:
## Description of Changes
- Adds two utility scripts under src/Translations to streamline localization maintenance:
  - check_translations.py : reports missing keys per language and prints a summary of counts.
  - insert_untranslated.py : inserts any missing keys at their original positions based on en-US order.

- Ensures values for non-English languages are prefixed with UNTRANSLATED followed by the original English text; English variants (en-*) are filled without the prefix.
- Preserves original file encoding (UTF-8/UTF-8 BOM/UTF-16) and newline style when writing changes.
- Verified all non-English languages have complete keys; English variants were filled using en-US values.
- How to run from src/Translations:
  - Check: python ./check_translations.py
  - Insert: python ./insert_untranslated.py

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Translation update

## Issues Resolved
- N/A


beeradmoore:
Hey @peter9811 ,
Thanks for the contribution. Unfortuantly I don't know python so I won't be able to maintain these going forwards and because of that I won't be bringing this in.

I already have tools for finding missing translations in the `extras/InNeedOfTranslationDetector/`. Additionally if you build the app in debug mode you can set the language to `LANG_HUNT` which will cause all translations to become "...". Any text that remains as words is not translated.
I also have a powershell script `main/extras/SortTranslations/` that I run regularly that will ensure translations are kept in the correct order. Though I don't think I do anything specific about line endings or BOM, that script should probably be updated. However that update should also make sure a fresh git clone, edit the file in Visual Studio, and export an updated translation from translation toolbox to maintain consistant line endings/BOM. 

Do those two tools do essentially what your check script does?

As for copying text across, we don't want to do that. The inbuilt translator tool (Settings > Language > Open translator toolbox) relies on text being missing to help translators see and update missing texts. It does not matter for the keys being missing but that should only be an issue for someone editing in a plain text editor and not Visual Studio (though the preferred method is translation toolbox)

Issue #777, opened 2025-12-03
[BUG] Improve translation files consistancy


beeradmoore:
None of these matter to the final product until something else changes and they someday do. It also helps merge translation files in if everything is the same spacing, same line ending, etc.

## Comments removed from translations

I need to reproduce (it could have just been the user did it) but the translations submitted in #749 for #751 had the comment tags stripped out. This may be DLSS Swapper tooling doing it, and if it is we should try retain those as they  may be useful to future translators.

## Translation toolbox line spacing issues

It is removing some line spacing when it shouldn't. See #818

## Consistant line endings

~~Translation toolbox + sort.ps1 may change line endings. We should ensure all of these are consistant.~~
Fixed in #856

PR #818, opened 2026-01-06
Add remaining German translation (Version 1.2.3.1)

melgu:
## Description of Changes

Added missing German translations for version 1.2.3.1

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Translation update

### DIFF
--- a/src/Translations/de-DE/Resources.resw
+++ b/src/Translations/de-DE/Resources.resw
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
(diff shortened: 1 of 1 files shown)


beeradmoore (approved):

beeradmoore:
Thanks @melgu 
There was some funky whitespace changes up the top that I reverted. I know it would not have changed anything with how the file runs I wanted to keep it how it was. In future I think I might apply that whitespace change to all the translation files so they are all like that to prevent needing to do it again.

melgu:
I used the translation toolbox inside DLSS Swapper itself. Looks like there's a mismatch between what that tool outputs and what's the current preferred format. So I agree, it makes sense to update one or the other.

beeradmoore:
Translation toolbox also strips out `<comments>`. It + powershell sort script may also change line endings. Again, I don't think it matters to the product but I'd rather make them all match before I am chasing my tail in a few years wondering why some languages work and others are borked after .NET changes something where it then does matter 😂

I have adjusted #777 to focus on consistancy in these files and tools.

PR #830, opened 2026-01-16
Fixes an issue where closing a window while minimized won't yeet it to -32000 x/y pos

beeradmoore:
## Description of Changes

If you minmise a window Windows sets the X/Y position to -32000, If you then close the window and re-open DLSS Swapper we will try put it at that position for it to never be seen again.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Translation update

## Issues Resolved

#829

Issue #829, opened 2026-01-15
DLSS Swapper opens but doesn't let me access it


drencoss:
### Describe your issue

By that i mean after installing DLSS Swapper and using it for a few days say  i go to launch DLSS Swapper from either the Taskbar or Desktop shortcut it opens as i can see if i hover my mouse over the icon in the Taskbar that is has opened aswell as it showing up in the Task Manager, but when i go to click on it to use it and have it go fullscreen (check screenshot) nothing happens it doesn't go Fullscreen so im stuck with out being able to access and use this.

But if i delete the json folder from C:\Users\AppData\Local\DLSS Swapper it will allow me to access and use it.

<img width="707" height="511" alt="Image" src="https://github.com/user-attachments/assets/b1fd5b74-6f96-4914-8461-cd249d5b8056" />

beeradmoore:
Hey @drencoss , do you happen to change your screen configuration sometimes? I am wondering if it is opening on a screen that no longer exists. We store the last used position/size in the app settings file and then when we open it again we restore back to that position. That would be why you discovered deleting `AppData\Local\DLSS Swapper\` fixed it.

Have you been using DLSS Swapper for long? In a recent change in v1.2.3.0 a slight change was made to windowing but I didn't think anything there would break it. Have you seen this issue longer than 2 weeks ago or less than that?

If you could, next time it is in a broken state can you please go to `AppData\Local\DLSS Swapper\json\settings.json` and upload that file here so I can look at if the data went bad inside it.

As for fixing it, pressing win+left or win+right on the keyboard may move it into view. Ideally I can find out what is wrong and fix the broken code.

andressergiox:
doesn't open for me after nvidia update to DLSS 310.5.2, stays in the taskbar, i can see the preview but don't open, tested without network, admin, etc did many things and nothing...

drencoss:
> Hey [@drencoss](https://github.com/drencoss) , do you happen to change your screen configuration sometimes? I am wondering if it is opening on a screen that no longer exists. We store the last used position/size in the app settings file and then when we open it again we restore back to that position. That would be why you discovered deleting `AppData\Local\DLSS Swapper\` fixed it.
> 
> Have you been using DLSS Swapper for long? In a recent change in v1.2.3.0 a slight change was made to windowing but I didn't think anything there would break it. Have you seen this issue longer than 2 weeks ago or less than that?
> 
[...]

PR #838, opened 2026-01-27
Removed magic numbers and added fix to prevent erroneous window position being set when closing app from minimized state.

andyeder:
## Description of Changes

Due to a missed edge case (sorry!) in my original window position/state persistence code, it was possible for the application window to warp to position (x: -32000, y: -32000) when closed from a minimized state. @beeradmoore kindly located, and resolved, the issue.

This takes that fix a step further by removing the "-32000" magic numbers ("where windows go to die") and by fixing the logic check to prevent erroneous values being written to the settings file when the application has entered either a minimized OR maximized state. Other than removing the magic numbers, it is reduced to a single line fix.

Basically, preventing the app window position from being tracked/saved when entering into a minimized/maximized state.

## Type of Change

_(Wasn't sure whether to class it as a Bug fix or New feature!)_

- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Translation update

## Issues Resolved

This is a follow-on from #830, which addressed #829.

### DIFF
--- a/src/Data/WindowPositionRect.cs
+++ b/src/Data/WindowPositionRect.cs
@@ -35,6 +35,18 @@ public class WindowPositionRect
         Width = other.Width;
         Height = other.Height;
         State = other.State;
+
+        // LEGACY: Restore broken windows to correct positions.
+        // -32000 is some magic number were windows go to die.
+        if (X == -32000)
+        {
+            X = 0;
+        }
+
+        if (Y == -32000)
+        {
+            Y = 0;
+        }
     }
 
     public WindowPositionRect(int x, int y, int width, int height)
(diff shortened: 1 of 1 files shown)


beeradmoore:
I'll review this fully when I have more time. 

I am torn on if we should add the removed `-32000` checks back in (at least in the part around loading, when we set it we can ignore it). The idea was if someone has a broken app and updates to the new app it will fix it automatically for them. That fix has been released for a week, but I think a lot of people don't regularly update things as much as I would. This is also something I wish I had analytics insight into.

andyeder:
No problem, and in light of the fact that there might be a lot of folks who don't update regularly... yeah, you have a good point. The -32000 checks certainly won't do any harm and - agreed - the benefit of having the check, at least on the loading section, could indeed bring a user's window back to the visible screen area!

beeradmoore:
Thanks @andyeder . I restored just two of the `-32000` checks. The two that happen at launch. This should be enough.

PR #883, opened 2026-04-26
Feature/streamline sdk update

RankFTW:
## Description of Changes

Adds Streamline SDK DLL management to DLSS Swapper. Unlike the existing per-DLL swap workflow, Streamline operates as a batch update — all Streamline DLLs in a game directory are updated to the latest SDK version at once, sourced from the NVIDIA-RTX/Streamline GitHub releases.

**What's included:**
- **StreamlineManager** singleton that fetches, caches, and stages the latest Streamline SDK release from GitHub
- **StreamlineUpdater** for batch update/restore of Streamline DLLs with backup (.dlsss) and rollback support
- **Game detection** — ProcessGame() now detects sl.*.dll files and tracks them as Streamline game assets
- **UI section** in the game detail dialog (between DLSS FG and FSR) showing current version, Update and Restore buttons
- **One-time migration** to re-scan existing games for Streamline DLLs on first launch
- **Real-time UI refresh** after update/restore operations
- **Property-based tests** covering DLL classification, version comparison, update intersection, backup path round-trip, and restore targets

**Design decisions:**
- No manifest integration — Streamline is a fundamentally different model (single GitHub ZIP vs per-version manifest), so a dedicated manager avoids polluting the existing architecture
- Reuses existing patterns: .dlsss backup convention, GameAsset/GameHistory recording, relay commands, singleton managers
- Version determined by sl.interposer.dll file metadata

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Translation update


beeradmoore:
Hey @RankFTW ,
Thanks for this. I'll need a big slot of time to go over this on the weekend in detail.

A few questions off the bat.

1. why was that one streamline file used in particular for version detection?

2. Is there any use to updating a single SL DLL and not all (or to downgrade a single one). This sounds like (haven't checked the code yet) that a single version is used.

My plan was going to allow all but bundle them all as "DLL Set". Updating all DLSS (Sr+fg+rr) would be a single set but allow people to modify each one individually. Same goes for FSR and direct storage, etc.



Only other comment so far is that at some point I will change streamline to be part of manifest.json with the option to import on import screen like the recently added NVIDIA import.

RankFTW:
Hey man, thanks for getting back to me.

1. the sl.interposer.dll is the main dll file for streamline so everything
else goes through that and it's the first one to load when streamline is
loaded
2. so right now it is implemented to update all the games installed
streamline files to be the same version, generally games use the same file
version for all their streamline files and tbh mixing streamline file
version is never something that i've tried

I did look at having different streamline versions available in the library
section but without access to the DLSS Swapper manifest it wasn't something
that I would be able to do easily.

So far I've tested it on a few games and it works well, swapping all the
streamline dlls installed for the newest version while backing up the
original files with the .dlsss extension.

If you have any more questions man just let me know :)

beeradmoore:
Thanks for the fast answer. That makes sense why that one. Are you ok if I commit into this PR?

beeradmoore:
The manifest file is generated by [this tool](https://github.com/beeradmoore/dlss-swapper-manifest-builder). It can't really run standalone to DLSS Swapper as the download links would still point to my server.

When I run the tool it creates some folders in `generated_files`, will download the entire existing DLSS Swapper library (though this can probably be removed) and then looks in SDK to bring in SDK zips. It also has a place for individual zipped DLLs that can be sourced from games.

I can probably on this weekend get it to bring in Streamline SDKs and seperate them as individual DLLs as well as have a list to make a DLL set, and then have that manifest + all files uploaded.

RankFTW:
"Are you ok if I commit into this PR?"

Oh absolutely man, go ahead. DLSS Swapper is an amazing tool. Streamline
updating was like the one thing missing imo. Do whatever you want with this
now, it's yours :)

beeradmoore:
Hey @RankFTW , Just an update on this and keeping you in the loop. I was looking at this to integrate the NVIDIA OTA servers for this as well, but any Streamline download from them does no thave `sl.interposer.dll` in it. 

I wonder why that is, or better yet, if it is an issue if it does not update with everything else. I'll let you know how that goes.

LepaBrenaa:
Any ETA on this being merged? This has been missing for a long time now

beeradmoore:
Still working on the backend so this can be updated to work like everything else.

I can see about getting an unsigned build out of this branch but it won't be what we run with for release so things may change when you swap to final build

beeradmoore:
I am going to close this issue off. Now that I have my updated manifest with Streamline, FSR4, and DirectStorage it does not appear that much of this will be re-usable for that update as this PR changes how DLLs are sourced and used for one single library.

Issue #157, opened 2023-09-01
[Feature request] Select button or DLSS auto update


HarryKleins:
Good morning,

Would it be possible to add a button to select a list of games and update the DLSS at the same time?

Because at the moment I have to update my games DLSS manually one by one. I have over 50 DLSS compatible ones.

THANK YOU

beeradmoore:
I don't know how I didnt see and reply to this earlier.

It is difficult to do. We'd almost need a UI where it is a list of games on the left and you checkbox each one that you want it to do and right is the new DLL picker from [DLSS Swapper v1.1](https://github.com/beeradmoore/dlss-swapper/releases/tag/v1.1) so you can update DLSS FG, RR, FSR, XeSS all in one go.

beeradmoore:
Will keep the thought of bulk updates open in my head and see what solution can work.

> Also, the latest DLSS is named 310.01. Shouldn't it be 3.10.1 instead?

DLSS 310.1 is DLSS 4. We display the version from the property on the DLL, DLSS 4 is NVIDIAs marketing term. FSR 3.1 is in a similar boat where the DLLs are labelled like 1.1.32189. 

We can add additional labels to display next to them which I might have to do. I just don't know how we call DLSS 4 vs DLSS 4.1 (future) if NVIDIA doesn't expose this to us.

> Finally, the latest version no longer detects Ubisoft games, whereas in version 3.8, my games were displayed correctly.

I will double check this, I don't believe I have Ubisoft Connect installed anymore. If they updated their stuff it would prevent us loading them in.

Are you able to grab [v1.0.5 portable](https://github.com/beeradmoore/dlss-swapper/releases/tag/v1.0.5.0) and run it (just so it doesn't interfere with your installed version) and let me know if it detects Ubisoft games?

> Additionally, the EA, Battle.net, GOG, Rockstar Games, and Rio client platforms do not seem to be supported. Is this planned for a future update?

GOG is added (also offline installers are detected without GOG Galaxy). Is it broken as well?
What is Rio?

Issue #71 is the master list where we link to the other issues for monitoring those and see what activity is like. If you want any of those others supported feel free to go give them a 👍

HarryKleins:
Good evening,

I’ve finally tested the portable version 1.1.1.0, and everything works perfectly: GOG, UBISOFT, and even Xbox Connect are all detected. I uninstalled the installable version 1.1.1.0, reinstalled it, and that resolved the issue.

[...]

PR #913, opened 2026-07-19
Add batch DLL swap: select multiple games and apply in one pass

RafaelHGOliveira:
## Description of Changes

Adds a selection mode to the games grid so several games can be updated in one operation, instead of opening each game individually.

### Flow

**1.** `Select` in the toolbar enters selection mode — cards become checkable and a bar appears with the selection count, `Select all`, `Apply DLL`, `Clear` and `Done`.

Games grid in selection mode with six games selected

**2.** `Apply DLL` opens a picker listing every DLL type, each with a version dropdown, plus a preset dropdown for the three DLSS types. Every row defaults to *Don't change*, so nothing is touched unless it is explicitly chosen.

Apply DLL dialog with a version and preset dropdown per DLL type

**3.** Applying runs sequentially behind a progress dialog, then shows a summary: how many changes were made across how many games, how many were skipped, how many errored, and a separate count for changes that did not apply because a game does not ship that DLL type.

Batch swap summary listing each change that was applied

### Notable details

- The picker is game-agnostic — it offers what is downloaded locally, and per-game compatibility is resolved at apply time rather than by intersecting the selection up front. A game that cannot take a change is reported, not silently dropped.
- Results distinguish three outcomes: *swapped*, *skipped* (something the user may want to act on) and *error*. "Game does not use that DLL type" is collapsed into a single count so it does not bury the actionable lines.
- Failures that need elevation are aggregated into one "relaunch as admin" message rather than one per game.
- All user-facing strings go through `.resw` / `ResourceHelper`. **Only `en-US` is populated in this PR** — the other locales fall back until translated.
- No changes to `GameAssetType` numbering, the database schema, or the manifest.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Issues Resolved

Related to #157.

The design sketched in that issue is **not followed exactly** here — the differences I am aware of:

- selection is not persisted across app restarts
- no per-game opt-out inside the picker once games are selected
- `Select all` is a plain toggle rather than tri-state
- no bulk "reset to original DLL"

Happy to rework any of these, or to split the PR up, if the design in #157 is the direction you want to take. Opening it as a draft so there is something concrete to react to.

### DIFF
--- a/src/Pages/GameGridPage.xaml
+++ b/src/Pages/GameGridPage.xaml
@@ -97,6 +97,20 @@
 
             <GridView PointerWheelChanged="MainGridView_PointerWheelChanged" ManipulationMode="None" x:Name="MainGridView" ItemsSource="{x:Bind CurrentCollectionView, Mode=OneWay}" SelectionMode="None" IsItemClickEnabled="True" ItemClick="GridAndListView_ItemClick" Padding="20"  >
 
+                <GridView.Resources>
+                    <!-- White selection checkmark backdrop for contrast against varied cover art (default uses the accent color). -->
+                    <SolidColorBrush x:Key="GridViewItemCheckBoxSelectedBrush" Color="White" />
+                    <SolidColorBrush x:Key="GridViewItemCheckBoxSelectedPointerOverBrush" Color="White" />
+                    <SolidColorBrush x:Key="GridViewItemCheckBoxSelectedPressedBrush" Color="#DDDDDD" />
+                    <SolidColorBrush x:Key="GridViewItemCheckBrush" Color="Black" />
+
+                    <!-- The item container draws behind the cover art, so its selected background only
+                         shows through the padding added below, reading as an accent frame. -->
+                    <StaticResource x:Key="GridViewItemBackgroundSelected" ResourceKey="AccentFillColorDefaultBrush" />
+                    <StaticResource x:Key="GridViewItemBackgroundSelectedPointerOver" ResourceKey="AccentFillColorSecondaryBrush" />
+                    <StaticResource x:Key="GridViewItemBackgroundSelectedPressed" ResourceKey="AccentFillColorTertiaryBrush" />
+                </GridView.Resources>
+
                 <GridView.GroupStyle>
                     <GroupStyle HidesIfEmpty="True">
                         <GroupStyle.HeaderTemplate>
@@ -115,6 +129,10 @@
                         <Setter Property="VerticalContentAlignment" Value="Stretch"/>
                         <Setter Property="Margin" Value="8"/>
                         <Setter Property="BorderThickness" Value="0"/>
+                        <!-- Padding + matching corner radius are what make the selected background
+                             visible as a frame around the 8px-rounded card content. -->
+                        <Setter Property="Padding" Value="3"/>
+                        <Setter Property="CornerRadius" Value="11"/>
                     </Style>
                 </GridView.ItemContainerStyle>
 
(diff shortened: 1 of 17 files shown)


beeradmoore:
Hey @RafaelHGOliveira, thanks for pushing this. I just wanted to let you know I see this PR, I just have not had time to review it yet.

lguerard:
Looking forward for that PR or a similar solution to be added to the amazing software !

src/Pages/GameGridPage.xaml
```diff
@@ -231,22 +249,27 @@
...
+                <AppBarButton Label="{x:Bind ViewModel.TranslationProperties.RefreshText, Mode=OneWay}" Command="{x:Bind ViewModel.RefreshGamesButtonCommand}" IsEnabled="{x:Bind ViewModel.CanRefresh, Mode=OneWay}">
                     <AppBarButton.Content>
                         <FontIcon Glyph="&#xE72C;" />
                     </AppBarButton.Content>
                 </AppBarButton>
-                <AppBarButton Icon="Filter" Label="{x:Bind ViewModel.TranslationProperties.FilterText, Mode=OneWay}" Command="{x:Bind ViewModel.FilterGamesButtonCommand}" IsEnabled="{x:Bind ViewModel.IsGameListLoading, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}">
+                <AppBarButton Label="{x:Bind ViewModel.TranslationProperties.SelectText, Mode=OneWay}" Command="{x:Bind ViewModel.ToggleSelectionModeCommand}" IsEnabled="{x:Bind ViewModel.IsGameListLoading, Converter={StaticResource BoolNegationConverter}, Mode=OneWay}">
```
beeradmoore, line 262: I think "Select" is not clear what it is doing. "Bulk Update" and "Bulk Swap" may be a better fit, unsure if they will be too long though. (EDIT: Batch is another word to throw in there, but I am not sure if gamers understand what "batch update" will do as much as developers would)

src/Pages/GameGridPage.xaml
```diff
@@ -267,6 +290,25 @@
...
+                VerticalAlignment="Bottom"
+                HorizontalAlignment="Center"
+                Margin="0,0,0,24"
+                Padding="16,8"
+                CornerRadius="8"
+                BorderThickness="1"
+                Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
+                BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
+                Visibility="{x:Bind ViewModel.IsSelectionMode, Mode=OneWay, Converter={StaticResource BoolToVisibilityConverter}}">
+            <StackPanel Orientation="Horizontal" Spacing="12">
```
beeradmoore, line 304: There is something about the buttons that isn't sitting right. I am not sure if these comments make it right but these are the things that call out to me.

"X selected" - great, and great position.
"Select all" - great, and great position.
"Apply DLL" - I like this as a primary button because it is the intended next button to click. I think there could be better wording. "Apply" implies clicking it will apply something but the user hasn't seen what they are applying (because they need to go here first to select it). Maybe something like "Select DLLs" or "Select Versions" or something? Open to other ideas if you have them.
"Clear" I think this should be "Select None" and should be moved left one. That is unless you wanted to swap "select all" to toggle to "select none" after they press it? Two buttons for these actions should be fine though.
"Done" - I think can be "Close", unsure if this shoudl be text, or the glyph for close.

beeradmoore (changes requested):
Hey @RafaelHGOliveira , apologies for the long time to review. I still have not had time to do a full review yet but have left some notes on what I have seen so far. I have not run locally yet but looking at the screenshots I think this method will work. It is not what I had invisioned in #157, however I think this has many upsides (assuming filter + search will work while bulk editing) and may be preferred.

I had an idea from when you said these two things:

> Every row defaults to Don't change, so nothing is touched unless it is explicitly chosen.

> no bulk "reset to original DLL"

To add support for that could there be a row in there after "Don't Change" for "Restore original" (or "Restore original DLL") so that functinality could exist?


These are the pages I have not reviewed yet:
- src/Pages/GameGridPageModel.cs
- src/UserControls/BatchDllPickerControl.xaml
- src/UserControls/BatchDllPickerControlModel.cs
- src/UserControls/BatchDllRowModel.cs
- src/UserControls/BatchSwapSummaryControl.xaml
- src/UserControls/BatchSwapSummaryControlModel.cs

beeradmoore:
Also another question, does this select method work if we are in list mode and not grid mode for the games?

I would love to have this in before the v1.3 release with Streamline, DirectStorage, FSR4, etc. But I don't know when that is, and I don't know when you are free to look at the above. If it is it is, if it isn't it isn't. I will have to update this screen after that to add support for them. I think I'll be doing changes on the game page to clean up the DLL selection like adding grouping of technolgoies under tabs or something. I'll look at applying those same changes here so they both match.

RafaelHGOliveira:
Thanks for the review @beeradmoore! I've pushed a follow-up that addresses the notes so far:

**Toolbar / selection bar**
- "Select" is now **Bulk Swap**.
- The bar now reads: `X selected` · **Select all** · **Select none** · **Select DLLs** (primary) · **Close**.
  - "Select all" only adds the currently visible games (it no longer toggles to deselect).
  - "Select none" clears the whole selection, including games hidden by search/filter.
  - "Select DLLs" opens the picker; nothing is applied until you confirm there.
  - I went with "Close" as text rather than a glyph so it reads consistently with the other buttons — happy to switch to the glyph if you prefer.

**Filter + search while bulk editing**
- Search and the Filter dialog now stay enabled in selection mode (Add Game and the grid/list switch stay disabled).
- Selections survive search/filter changes: games that get hidden stay selected and are still counted, and their checks come back when they are visible again. Applying a filter keeps the current search text.
- With "group by library" on, a favourite appears twice (Favourites + its library). Selection is per game, so both appearances show the same state and it is counted once.

**Restore original DLL**
- Every DLL row now offers **Don't change → Restore original DLL → versions**. Rows for FSR/XeSS are always shown, so a restore is possible even when no version of that type is downloaded.
- A restore never triggers a download and uses the existing `ResetDllAsync` path. It can be combined with a preset change on the same row; each outcome is reported separately.
- In the summary: a game whose DLL has no `.dlsss` backup already uses the original, so it goes into the "not applicable" count; a backup without a matching game DLL is reported as an error that needs a manual repair.

**List mode**
Yes — selection mode works on whichever view is active (grid or list). Switching between them is disabled while selecting, but you can enter selection mode from either.

**Translations**
The summary title now says "Bulk swap" too, to match the toolbar. I also added pt-BR for all the new strings (I'm a native speaker); the other locales still fall back to English until translated.

**Screenshots**

**Toolbar** with the new Bulk Swap button:

image1


**Selection bar** in bulk swap mode:

image2


**DLL picker** with the Restore original DLL option:

image3


**Summary** after a bulk restore:

image4

beeradmoore:
> I went with "Close" as text rather than a glyph so it reads consistently with the other buttons — happy to switch to the glyph if you prefer.

I think this was a good call.

Everything else looks great. I love the extra bug fixes added in as well. I can't wait to find time to re-review and get this merged in. Thanks for the amazing work @RafaelHGOliveira

src/Pages/GameGridPage.xaml.cs
```diff
@@ -192,4 +195,134 @@ private void ClearSearchBox_Click(object sender, RoutedEventArgs e)
...
+        _viewSyncGeneration++;
+    }
+
+    // Called after CurrentCollectionView changes while selection mode is active.
+    // The binding applies the new ItemsSource asynchronously, so the visual
+    // selection is re-applied at low priority, after the list picked it up.
+    internal void ResyncVisualSelectionAfterViewChange()
+    {
+        var generation = _viewSyncGeneration;
+        var enqueued = DispatcherQueue.TryEnqueue(Microsoft.UI.Dispatching.DispatcherQueuePriority.Low, () =>
```
beeradmoore: Any particular reason this is `DispatcherQueuePriority.Low`? I don't recall bothering to set it in the past and I am unsure if I am doing something wrong.

RafaelHGOliveira: No good reason, it turns out. I picked Low while chasing the selection bug, thinking the list control had to finish handling the new ItemsSource before the selection could be re-applied, and never went back to check.

I tested it this time. In bulk swap mode I selected a few games, then searched, cleared, typed quickly and re-applied the filter over 15 rounds. The checks came back fine at Normal priority, and even when the resync ran synchronously. With the resync removed they disappear, so the test does catch the problem.

So you're not doing anything wrong by leaving it at the default. I've switched it to the default and fixed the comment above the method, which described the same wrong assumption (c627e08).

beeradmoore: All good. I only ask because I used this to then look at how I am using `DispatcherQueue` elsewhere. I made an issue (#948) for me to deal with later to unify all instances of `DispatcherQueue` with the `RunOnUIThread`. It was brought to my attention [here](https://github.com/beeradmoore/dlss-swapper/pull/933#issuecomment-5574901286) that it was not working as I thought in all situations. That's why I was wondering if I should be using `Low` elsewhere when I re-vamp the UI threading code.

Changing to `RunOnUIThread` is not required here for your code.

beeradmoore:
Code wise this is all approved. I just need to test it and we should be good to go.

Issue #928, opened 2026-09-03
[Feature request] Add a banner for update available


beeradmoore:
### Describe what this adds or solves

Only at one time will the app say there is an update. If the user skips this, or a bug happen and it does not show then the user may never know. 

### Describe the solution you'd like

Add a banner to the bottom of the app, or top of the games page or something like that so it can say "Update available: v1.2.7". Clicking it should present the update page as normal.

hammadk306:
Hi! I would love to work on this issue. Could you please assign it to me? Thanks!

beeradmoore:
It looks like a PR is already open for it

beeradmoore:
Apologies, that is to fix the popup. You can have a shot at it.

I was wondering if a status bar that has a green pill shaped thing in the bottom right would be good. It shouldn't be excessive and intrusive. Happy for you to have a play with it to see what works and what doesn't

PR #932, opened 2026-09-05
Fixed update prompt never showing for a newer version

### LINKED ISSUE #924: [BUG] Auto-Update does not appear to show an update
### Describe the bug

DLSS Swapper v1.2.6.1 is out but DLSS Swapper v1.2.5 does not prompt to update.

### Steps To Reproduce

1. Open DLSS Swapper
2. Wait for an update screen.

### Expected behavior

Popup saying about a new update available.

### Logs

_No response_

### Screenshots

_No response_

### Additional context

This was first mentioned [here](https://github.com/beeradmoore/dlss-swapper/issues/912#issuecomment-5404763819).

### DLSS Swapper version

v1.2.5

### Windows version

25H2

ethanstoner:
## Description of Changes

`GitHubUpdater.HasPromptedBefore` has its comparison the wrong way round, so once the app has prompted about any release it never prompts again.

The method answers "have we already nagged the user about this version?". `MainWindow.xaml.cs:379` only shows the dialog when it returns `false`. As written it returns `false` when the available release is **older** than `Settings.Instance.LastPromptWasForVersion`, and `true` for every version **newer** than it, which is backwards:

```csharp
else if (thisVersion < lastVersionPromptedFor)
{
    return false;
}
```

That is the case in #924. A user on v1.2.5 already has `LastPromptWasForVersion` set to v1.2.5.0 from the previous prompt. v1.2.6.1 is greater than that, so the check reports "already prompted" and the dialog is skipped. The `LastPromptWasForVersion == 0` early return is why it still works exactly once, on a fresh install.

This is a one character change, `<` to `>`.

I also checked a second possible cause before settling on this one. `GitHubRelease.GetVersionNumber` returns 0 unless the release name starts with `v`, which would break the check a different way, but the API really does return `name=v1.2.6.1`, so parsing is fine.

### Evidence

I copied `GetVersionNumber` and `HasPromptedBefore` into a standalone console app and ran the real release names from the API. With `LastPromptWasForVersion` at v1.2.5.0:

| latest release | current: prompts? | patched: prompts? | should prompt? |
| --- | --- | --- | --- |
| v1.2.4.0 | yes | no | no |
| v1.2.5.0 | no | no | no |
| v1.2.6.1 | no | yes | yes |
| v1.3.0.0 | no | yes | yes |
| v2.0.0.0 | no | yes | yes |

The current code is wrong in four of the five cases, including every newer version. With the operator flipped all five are correct, and the first run case (`LastPromptWasForVersion == 0`) still prompts either way.

Builds clean, 0 warnings and 0 errors.

One thing I have not verified: the dialog actually appearing in a running app. That needs a real prompt cycle against a stale `LastPromptWasForVersion` and a newer published release, which I could not set up locally. The logic above is verified, the end to end UI path is not. Happy to test it another way if you would like.

This should also cover most of what #928 is about, since that describes the same "only prompts once" behaviour, though a persistent banner would still be a useful addition on top.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Translation update

## Issues Resolved

#924

### DIFF
--- a/src/Data/GitHub/GitHubUpdater.cs
+++ b/src/Data/GitHub/GitHubUpdater.cs
@@ -158,7 +158,7 @@ internal class GitHubUpdater
         {
             return false;
         }
-        else if (thisVersion < lastVersionPromptedFor)
+        else if (thisVersion > lastVersionPromptedFor)
         {
             return false;
         }


beeradmoore:
> so once the app has prompted about any release it never prompts again.

I will double check the rest of the logic but this is intentional (though labelled incorrectly). If a user is on 1.2.5 and 1.2.6 comes out it should prompt only once. The intent being that the user does not get a prompt every time.

You have raised some interesting points though so I'd love to test it out. All of my tests have been on apps with fresh install so I could be deceiving results. 

Thinking about the update feature maybe this logic should be changed so the popup shows every time but it also has a "[ ] Don't prompt me for this version again". The internal checks mentioned here should then check if the upcoming version matches the one they clicked "don't prompt..." then we don't display the popup. Until then they get the popup every single launch.

ethanstoner:
That intent is already what the write back gives you, and it survives the fix.

I simulated the full lifecycle rather than just the comparison: `HasPromptedBefore` (`GitHubUpdater.cs:152-167`) plus the `LastPromptWasForVersion` write back in `DisplayNewUpdateDialog` (`GitHubUpdater.cs:171-176`), starting from a user on v1.2.5.0 who has already been prompted for it, then two app launches each across v1.2.6.1 and v1.3.0.0:

- shipped: silent, silent, silent, silent. 0 prompts across both releases.
- with the change: PROMPT, silent, PROMPT, silent. 2 prompts, one per release, never twice for the same one.

So the "not every time" behaviour you want comes from the write back, not from the `<`. The comparison only decides whether a newer release is eligible at all, and today it answers no for every release above the one already recorded.

On the checkbox design, that is better than either version, but it needs a new setting, XAML and localisation, so it is a bigger change than this one. Do you want to land this as the correctness fix now and do the checkbox separately, or hold #932 and go straight to the checkbox?

beeradmoore:
Thanks for the additional testing. I'll double verify it when I can. I think this is a pre-requisite for the checkbox functionality to work right so I'll be happy to merge this in rather than do the checkbox work and make the bool change on that branch.

I don't know if you'd like to give the checkbox a shot, you are more than welcome to. If not I can do it. 

Even if I do it I could apply this boolean change on that branch, however work went in the issue here to diagnose the issue and find a solution. Contributors list should reflect that you did indeed contribute to a fix.

beeradmoore:
I was able to verify in a differnet way.

Whenever GitHubUpdater would return a version I set it to `var versionNumber = new Version(1, 2, 6, 0).GetVersionNumber();` and then deleted `settings.json`. Launching the app I got a popup. Bumping version that I added to 1.2.6.1 I did not get the popup.

Using your PR and doing the same test I was able to get it to pop the second time.

beeradmoore (approved):
Thanks for discovering and resolving this issue @ethanstoner

PR #933, opened 2026-09-06
Fix Steam scan crash from updating UI-bound properties off the UI thread

### LINKED ISSUE #909: [BUG] Some game detection fails - updating UI from the background thread
### Describe the bug

I have had to add Ratchet and Clank rift apart, jedi survivor, star wars outlaws, hellblade II, and I think a few others.

I saw that it said this might be a bug and to make an issue on Github so here i am.

Image

Image
Image
Image

### Steps To Reproduce

Open app and games weren't there

### Expected behavior

The expectation was that they would automatically pull into DLSS. 

Honestly though i don't find this to be that big of a deal. i think manually assigning isn't that hard and i still love this app.

### Logs

_No response_

### Screenshots

_No response_

### Additional context

Windows 11 Pro
Device name	Link
Processor	AMD Ryzen 9 9900X 12-Core Processor (4.40 GHz)
Installed RAM	96.0 GB (93.6 GB usable)
Graphics card	AMD Radeon RX 7900 XTX (24 GB)
AMD Radeon(TM) Graphics (2 GB)
Storage	3.96 TB of 29.15 TB used
System type	64-bit operating system, x64-based processor

### DLSS Swapper version

1.2.5

### Windows version

Version 25H2 (26200.8655)

Swissola:
## Description of Changes

Fixes the scan that never finishes / spinner stuck forever bug described in #909.

Steam, Battle.net, Epic Games Store, Ubisoft Connect and EA App all set UI-bound `Game` properties (`Title`, `InstallPath`, `StateFlags`, etc.) after execution has already moved off the UI thread (via an earlier `ConfigureAwait(false)`, or for EA App inside a `Parallel.ForEachAsync` body, which never runs on the UI thread at all). When a bound control actually needed updating, WinUI threw `COMException (0x8001010E)` / `RPC_E_WRONGTHREAD`, which was never caught. That aborted the scan for every remaining game in that library and left `GameGridPageModel.InitialLoadAsync` unable to reach the code that clears the loading spinner, leaving it stuck indefinitely - matching the log signature in #909's thread exactly.

Three commits:

1. **Fix RPC_E_WRONGTHREAD crash in game library scanners** - marshal the UI-bound property writes back onto the UI thread with `App.CurrentApp.RunOnUIThreadAsync` before continuing, in the five library scanners that were actually vulnerable (verified by tracing which ones had a context-dropping `await` ahead of the property writes; GOG, Xbox and ManuallyAdded don't have this issue and are untouched).
2. **Ensure game grid loading spinner always clears on scan failure** - defence in depth: wrap `InitialLoadAsync`'s loads in `try/finally` so the spinner can't get permanently stuck again from some other future exception, while still letting `GameGridPage.xaml.cs`'s existing `SafeFireAndForget` logging see the exception.
3. **Limit concurrent game processing to avoid a startup I/O/UI flood** - every game with no previously known DLLs gets re-queued for a full recursive directory scan on every launch (pre-existing behaviour); previously this rarely ran to completion because of the crash above. Now that scans actually finish, an unthrottled library would fire off one concurrent scan (and UI-thread update) per not-yet-processed game with no cap at all. Capped concurrent `ProcessGame` executions with a `SemaphoreSlim` so this doesn't overwhelm disk I/O or the UI thread. The chosen limit (4) is still under discussion in review, see the PR comments.

#909 was reported against a 2-drive library (Ryzen 9 9900X, 12c/24t). @beeradmoore independently re-tested this fix against his own real ~60-game library spread across several SSDs (5950X, 16c/32t): the scan that previously never completed now finishes reliably, averaging 12.2s with this PR's `SemaphoreSlim(4)` versus 23.8s before it. See the PR review thread for the full concurrency-tuning discussion, including further local benchmarking, that followed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Translation update

## Issues Resolved

#909

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014mYanCg6fAVVe6JGkAjTRw

### DIFF
--- a/src/Data/Game.cs
+++ b/src/Data/Game.cs
@@ -234,6 +234,11 @@ public abstract partial class Game : ObservableObject, IComparable<Game>, IEquat
         };
     }
 
+    // Bounds how many games can be scanned for DLLs/covers at once. Every game with no previously known
+    // DLLs is re-queued for processing on every launch, so without a limit a large library fires off
+    // hundreds of concurrent recursive directory scans and UI-thread updates at startup.
+    static readonly SemaphoreSlim processGameSemaphore = new SemaphoreSlim(4);
+
     /// <summary>
     /// Detects DLSS and updates cover image.
     /// </summary>
@@ -268,6 +273,8 @@ public abstract partial class Game : ObservableObject, IComparable<Game>, IEquat
 
         ThreadPool.QueueUserWorkItem(async (stateInfo) =>
         {
+            await processGameSemaphore.WaitAsync().ConfigureAwait(false);
+
             var newHasSwappableItems = false;
 
             try
@@ -564,6 +571,8 @@ public abstract partial class Game : ObservableObject, IComparable<Game>, IEquat
             }
             finally
             {
+                processGameSemaphore.Release();
+
                 // Now update all the data on the UI therad.
                 await App.CurrentApp.RunOnUIThreadAsync(async () =>
                 {
--- a/src/Data/BattleNet/BattleNetLibrary.cs
+++ b/src/Data/BattleNet/BattleNetLibrary.cs
@@ -263,44 +263,52 @@ internal partial class BattleNetLibrary : IGameLibrary
                 var cachedGame = GameManager.Instance.GetGame<BattleNetGame>(gameId);
                 var activeGame = cachedGame ?? new BattleNetGame(gameId);
 
-                if (_knownGames.TryGetValue(product.Uid, out var battleNetLauncherGame))
+                // These are UI-bound properties. Once the SaveToDatabaseAsync().ConfigureAwait(false) below has run for
+                // an earlier game in this loop we are no longer on the UI thread, so these must be marshalled explicitly
+                // or WinUI throws RPC_E_WRONGTHREAD when a bound control is updated from a background thread.
+                await App.CurrentApp.RunOnUIThreadAsync(() =>
                 {
-                    activeGame.Title = battleNetLauncherGame.Name;
-                    activeGame.LauncherId = battleNetLauncherGame.LauncherId;
-                }
-                else
-                {
-                    Logger.Error($"Battle.Net game title not found for UID ({product.Uid}) in install path ({product.Settings.InstallPath}).");
-
-                    if (string.IsNullOrWhiteSpace(product.Settings.InstallPath))
+                    if (_knownGames.TryGetValue(product.Uid, out var battleNetLauncherGame))
                     {
-                        activeGame.Title = product.Uid;
+                        activeGame.Title = battleNetLauncherGame.Name;
+                        activeGame.LauncherId = battleNetLauncherGame.LauncherId;
                     }
                     else
                     {
-                        var directoryInfo = new DirectoryInfo(product.Settings.InstallPath);
-                        activeGame.Title = directoryInfo.Name;
-                    }
-                }
+                        Logger.Error($"Battle.Net game title not found for UID ({product.Uid}) in install path ({product.Settings.InstallPath}).");
 
-                if (installedAggregates.TryGetValue(product.ProductCode, out var aggregate))
-                {
-                    // If title isn't set, try use it from the aggregates.
-                    if (string.IsNullOrWhiteSpace(activeGame.Title))
+                        if (string.IsNullOrWhiteSpace(product.Settings.InstallPath))
+                        {
+                            activeGame.Title = product.Uid;
+                        }
+                        else
+                        {
+                            var directoryInfo = new DirectoryInfo(product.Settings.InstallPath);
+                            activeGame.Title = directoryInfo.Name;
+                        }
+                    }
+
+                    if (installedAggregates.TryGetValue(product.ProductCode, out var aggregate))
                     {
-                        activeGame.Title = aggregate.Name;
+                        // If title isn't set, try use it from the aggregates.
+                        if (string.IsNullOrWhiteSpace(activeGame.Title))
+                        {
+                            activeGame.Title = aggregate.Name;
+                        }
+
+                        // Set the cover photo.
+                        activeGame.RemoteCoverImage = aggregate.LogoArtUri;
+                    }
+                    else
+                    {
+                        Logger.Error($"Battle.Net game aggregate not found for ProductCode ({product.ProductCode}).");
                     }
 
-                    // Set the cover photo.
-                    activeGame.RemoteCoverImage = aggregate.LogoArtUri;
-                }
-                else
-                {
-                    Logger.Error($"Battle.Net game aggregate not found for ProductCode ({product.ProductCode}).");
-                }
+                    activeGame.InstallPath = PathHelpers.NormalizePath(gamePath);
+                    activeGame.StatePlayable = product.CachedProductState.BaseProductState.Playable;
 
-                activeGame.InstallPath = PathHelpers.NormalizePath(gamePath);
-                activeGame.StatePlayable = product.CachedProductState.BaseProductState.Playable;
+                    return Task.CompletedTask;
+                }).ConfigureAwait(false);
 
                 if (activeGame.IsInIgnoredPath())
                 {
--- a/src/Data/EAApp/EAAppLibrary.cs
+++ b/src/Data/EAApp/EAAppLibrary.cs
@@ -176,9 +176,18 @@ internal class EAAppLibrary : IGameLibrary
                                     var cachedGame = GameManager.Instance.GetGame<EAAppGame>(contentId);
                                     var activeGame = cachedGame ?? new EAAppGame(contentId);
 
-                                    activeGame.Title = name;
-                                    activeGame.InstallPath = installPath;
-                                    activeGame.DisplayIconPath = programUninstallSubKey.GetValue("DisplayIcon")?.ToString()?.Trim('"') ?? string.Empty;
+                                    // These are UI-bound properties. Parallel.ForEachAsync above always schedules its
+                                    // work on the thread pool, never the UI thread, so these must be marshalled
+                                    // explicitly or WinUI throws RPC_E_WRONGTHREAD when a bound control is updated
+                                    // from a background thread.
+                                    await App.CurrentApp.RunOnUIThreadAsync(() =>
+                                    {
+                                        activeGame.Title = name;
+                                        activeGame.InstallPath = installPath;
+                                        activeGame.DisplayIconPath = programUninstallSubKey.GetValue("DisplayIcon")?.ToString()?.Trim('"') ?? string.Empty;
+
+                                        return Task.CompletedTask;
+                                    }).ConfigureAwait(false);
 
                                     if (activeGame.IsInIgnoredPath())
                                     {
--- a/src/Data/EpicGamesStore/EpicGamesStoreLibrary.cs
+++ b/src/Data/EpicGamesStore/EpicGamesStoreLibrary.cs
@@ -144,9 +144,18 @@ internal class EpicGamesStoreLibrary : IGameLibrary
 
                 var cachedGame = GameManager.Instance.GetGame<EpicGamesStoreGame>(manifest.CatalogItemId);
                 var activeGame = cachedGame ?? new EpicGamesStoreGame(manifest.CatalogItemId);
-                activeGame.RemoteHeaderImage = remoteHeaderUrl;
-                activeGame.Title = manifest.DisplayName; // TODO: Will this be a problem if the game is already loaded
-                activeGame.InstallPath = PathHelpers.NormalizePath(manifest.InstallLocation);
+
+                // These are UI-bound properties. The File.ReadAllTextAsync().ConfigureAwait(false) above has already
+                // taken us off the UI thread, so these must be marshalled explicitly or WinUI throws RPC_E_WRONGTHREAD
+                // when a bound control is updated from a background thread.
+                await App.CurrentApp.RunOnUIThreadAsync(() =>
+                {
+                    activeGame.RemoteHeaderImage = remoteHeaderUrl;
+                    activeGame.Title = manifest.DisplayName;
+                    activeGame.InstallPath = PathHelpers.NormalizePath(manifest.InstallLocation);
+
+                    return Task.CompletedTask;
+                }).ConfigureAwait(false);
 
                 if (activeGame.IsInIgnoredPath())
                 {
--- a/src/Data/Steam/SteamLibrary.cs
+++ b/src/Data/Steam/SteamLibrary.cs
@@ -220,15 +220,22 @@ internal partial class SteamLibrary : IGameLibrary
             var cachedGame = GameManager.Instance.GetGame<SteamGame>(game.PlatformId);
             var activeGame = cachedGame ?? game;
 
-            if (activeGame.IsHidden is null && _defaultHiddenGames.Contains(activeGame.PlatformId))
+            // These are UI-bound properties. Once the SaveToDatabaseAsync().ConfigureAwait(false) below has run for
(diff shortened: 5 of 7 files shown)


beeradmoore:
I could be wrong but I believe this will break things. Because that UI thread code dispatches its action the order of operations can go weird. It queues something to be done later, not waits for it to be done. eg. from this pseudocode, 


```csharp
await DispatcherQueue.EnqueueAsync(() => {
    Title = "Test";
}); 

if (Title == "Test") // this can be false
```

Swissola:
Fair worry in general, but I don't think it applies here, and I went and checked rather than just asserting it.

`DispatcherQueue.TryEnqueue` is the fire-and-forget one, queues the action and returns straight away, which is the failure mode your pseudocode is describing. But `RunOnUIThreadAsync` (in `App.xaml.cs`) wraps `DispatcherQueue.EnqueueAsync` from `CommunityToolkit.WinUI.Extensions`, and that's a different API. It wraps the queued delegate in a `TaskCompletionSource` and only completes that task once the delegate has actually run (or faulted, in which case the exception propagates back to the awaiter). I confirmed the `TaskCompletionSource` reference is present in the shipped `CommunityToolkit.WinUI.Extensions.dll` (8.2.251219), which matches the known implementation.

So the equivalent of your example, with the API actually in use, is:

```csharp
await App.CurrentApp.RunOnUIThreadAsync(() => { Title = "Test"; return Task.CompletedTask; });
if (Title == "Test") // always true, the await doesn't resume until the delegate has run
```

I checked every call site added in this PR (Steam, BattleNet, EAApp, EpicGamesStore, UbisoftConnect, plus the existing ones in `Game.cs` and `GitHubUpdater.cs`), and they all `await` it directly rather than firing it off, so the reads that follow (`IsInIgnoredPath()` and so on) are safe.

beeradmoore:
Oh interesting. Every other place I used `.TryEnqueue` instead of `.EnqueueAsync`. I use `.RunOnUIThread(` 51 times. It is the one that will have the negative behaviour like I said, and then `.RunOnUIThreadAsync(` 4 times.

I don't know if I ever knew this other method actually worked or if I had discovered the first one didn't and assumed the other didn't either. I checked through some usages of `RunOnUIThread` and see they could (and probably should) be `RunOnUIThreadAsync` (out of scope of this PR, but interesting to know)

src/Data/Game.cs
```diff
@@ -234,6 +234,11 @@ protected void SetID()
         };
     }
 
+    // Bounds how many games can be scanned for DLLs/covers at once. Every game with no previously known
+    // DLLs is re-queued for processing on every launch, so without a limit a large library fires off
+    // hundreds of concurrent recursive directory scans and UI-thread updates at startup.
+    static readonly SemaphoreSlim processGameSemaphore = new SemaphoreSlim(4);
```
beeradmoore, line 240: Is there any reason this is 4? 

Outside of the scope of this PR but my intent for loading was going to be having queues per disk drive and we throw games onto the queue depending where it is. Then a queue for HDD can process 1 game at a time, queue on SSD can process more (2, 4, more?). The idea beaing we don't saturate a specific drive and can scale up depending on performance of the drive. 

I added a timer into `LoadGamesAsync`.

```csharp
    public async Task LoadGamesAsync(bool forceNeedsProcessing = false)
    {
        var stopWatch = new Stopwatch();
        stopWatch.Start();
        ...

        stopWatch.Stop();

        Debug.WriteLine($"GAME LOAD TIME: {stopWatch.Elapsed.TotalSeconds}");
    }
````

For my current library of 60 games across several SSDs I would delete `bin/` and then build for `Debug_Portable` and see the total seconds this method took to load and took the average across 3 runs. This is on a 5950X (16 core, 32 thread)

| Test | Time (seconds)|
|--------|---------------| 
| Before this PR | 23.8 | 
| This PR | 12.2 |
| This PR - SemaphoreSlim(8) | 15.4 |
| This PR - SemaphoreSlim(2) | 7.3 |
| This PR - SemaphoreSlim(1) | 5.7 |

Seeing the result of this PR it was great to see it drop in half! Giving it more cores (but not too many) I expected more performance but it went down. Then dropping more and more I actually got more and more performance. I wonder if `SemaphoreSlim` is the right choice here if only having 1 thread load games at once would be better.

At this point the entire `ThreadPool.QueueUserWorkItem(` seems pointless and this should just run one at a time in a loop on the background thread.

Thoughts?

beeradmoore (changes requested):

src/Data/Game.cs
Swissola: I built a local harness to actually test this rather than guess. It mirrors Game.cs's pipeline as closely as I could manage outside the real app: `ThreadPool.QueueUserWorkItem` + `SemaphoreSlim`, a single-thread dispatcher with the same "only completes once the delegate has run" semantics as `RunOnUIThreadAsync`, a real SQLite connection behind a Nito.AsyncEx `AsyncLock` doing the same insert/delete pattern as `SaveToDatabaseAsync` and the `game_asset` writes, real `nvngx_dlss.dll` files (56-74MB, copied from games actually installed here) hashed with MD5 exactly like `LoadVersionAndHash`, and, once I suspected it mattered, the `ResizeCoverAsync` step too (ImageSharp decode + Lanczos5 downscale + PNG encode, held under the same semaphore slot as the hash, same as `Game.cs:562-564` sitting before the `finally` release).

Short version: I can't reproduce the inversion. On this machine (Ryzen 7 9800X3D, 16 threads, NVMe), every version of the pipeline I tried got faster with more concurrency, not slower:

| Games | Concurrency 1 | 2 | 4 | 8 |
|---|---|---|---|---|
| Full pipeline (hash+DB+dispatch) | 2.76s | 1.46s | 0.77s | 0.49s |
| Hash only, no DB/dispatch | 2.60s | 1.38s | 0.72s | 0.43s |
| Cover resize only (Lanczos5) | ~0.8s | ~0.75s | ~0.74s | ~0.75s |

That last row is flat rather than improving, but it never gets *worse* the way your table shows. I tried the DB-write-nested-in-`RunOnUIThreadAsync` thing I flagged in my last comment as a possible cause too (moved `SaveToDatabaseAsync` out from under the UI dispatch) and it made no measurable difference either way.

Worth being upfront: my first pass at this harness *did* show a dramatic false inversion, 20 games taking 210ms at concurrency 1 and ~2.7s at concurrency 2/4/8, identical regardless of what I changed. Turned out to be a bug in my own test code, not yours: I'd called the processing method directly instead of via `ThreadPool.QueueUserWorkItem`, so its synchronous prefix (the `Directory.GetFiles`/hash) ran inline on the calling thread whenever the semaphore had a free slot, quietly serialising everything behind it. Once I fixed the harness to actually go through `ThreadPool.QueueUserWorkItem` like `Game.cs` does, the false inversion disappeared. I mention it because it's the kind of mistake that's easy to make by hand too, and I want to be clear I'm not hand-waving past a result I don't like.

Two things I couldn't replicate that seem like the most plausible remaining explanations for what you're seeing:

1. **Your real SQLite database, mine was always empty.** Every run of my harness starts from a fresh `.db` file. Yours has presumably accumulated `game_history` rows across 60 games over a long time. If insert latency under `Database.Instance.Mutex` scales with table/journal size in a way an empty synthetic db won't show, more concurrent workers piling onto that mutex could genuinely get worse as the db grows, and I'd have no way to see that here.
2. **Whether the covers were already cached.** `shouldUpdatedCover` skips the resize for anything under 7 days old, but on a fresh `Debug_Portable` build (deleted `bin/`) every game likely re-resizes on the first run. Does the slowdown persist on a *second* consecutive run, where most covers should already be cached? If it goes away, that points at `ResizeCoverAsync` despite my isolated test of it not showing a regression, since 20 images might just not be enough to see whatever's happening at 60.

Also: were the numbers in your table single runs, or averaged? Worth ruling out ordinary run-to-run noise before chasing anything more subtle, since I only saw stable numbers once I averaged 3 repeats per config.

On the actual question, given I can't currently demonstrate more concurrency hurting anything, I'd hold off on ripping out `SemaphoreSlim` for a plain sequential loop. A sequential loop would only fix this by accident if the real cause is something that scales badly with concurrent access, and I haven't been able to show that's what's happening.

beeradmoore: > I would delete bin/ and then build for Debug_Portable and see the total seconds this method took to load and took the average across 3 runs

Average across 3 runs. Because I was using Debug_Portable it stores all of its local data in StoredData within its own directory. So deleting bin is clean DB and 0 image cache. My test was for testing explicitly for first launch as it is the worst launch.

I can try do an average once it has a cache to see if it replicates what you have.

My test was also testing the user experience, not trying to break it down per component to see where the slowdown is. Testing each component and trying to optimise and doing new things like moving to disk based queues, using MFT instead of listing file manually, etc, is a job for later down the track. 

The change you have is already a substantial improvement and I don't think it needs to be changed drastically for this to be merged in. I'm just curious why 4, why not scale on number of cores, why not scale on number of drives, or something similar. Thats why when I tested 1,2,4,8 I was surprised of the result 😂

I was also testing in debug mode. In theory release would potentially be faster, but I don't think it would scale differently per semaphore count.

Swissola: Reran everything in Debug config to rule that out too, since that was the last uncontrolled variable between our tests. Doesn't change anything: full pipeline scales the same way as it did in Release, 2.64s / 1.36s / 0.73s / 0.44s across concurrency 1/2/4/8, still faster with more concurrency, no inversion. The cover-resize path is flat under Debug as well, same as Release.

So both of my remaining theories are dead: you've confirmed your test was cold-DB/cold-cache like mine, and now Debug vs Release makes no difference here either. I'm out of things to try on my end without your hardware, your library, or the averaged warm-cache numbers you mentioned wanting to try.

This has been a good one to chase regardless, and appreciate you pushing back on it rather than just taking the PR as-is. If your warm-cache run turns up something concrete I'm happy to keep digging, but what to do with any of this is entirely your call, it's your codebase and your longer-term plan for the per-drive queueing anyway. Nice work catching the surprise in the first place.

Swissola: Update, and a partial retraction: I can now reproduce a version of this on the real code, not just my synthetic harness.

I pulled the real number for my own installed library straight from my production `dlss_swapper.db` rather than guess: 693 Steam + 3 GOG + 2 manually added = 698 games, across 4 drives (not the 5 I'd assumed from the PR description). Much bigger than the 60-game synthetic proxy I'd been using. Built Debug, added your exact `Stopwatch`/`GAME LOAD TIME` snippet to `LoadGamesAsync`, wiped the portable data for a clean cold-cache run each time, and swept `SemaphoreSlim` 1/2/4/8 with 3 repeats each against that real 698-game library:

| Concurrency | Avg GAME LOAD TIME |
|---|---|
| 1 | 2.82s |
| 2 | 2.82s |
| 4 | 3.03s |
| 8 | 3.44s |

That's clean and monotonic across all three repeats at each level, not noise. Same direction as your table, more concurrency, slower, just much smaller in magnitude here (+22% worst-to-best vs your +170%). So the effect is real on the actual code, my synthetic harness just wasn't a faithful enough proxy to catch it, probably because "20-60 synthetic games finishing in under a second" isn't the same regime as "698 real games with real cover art sustaining background contention for tens of seconds."

That said, my 9800X3D is a single-CCD part, 8 cores, one chiplet. Your 5950X is two 8-core CCDs joined over Infinity Fabric, and cross-CCD cache/memory access carries a real latency penalty, every version of this leans on shared synchronisation (`SemaphoreSlim`, `AsyncLock`, a single UI dispatcher thread) that all workers contend on, and that penalty can only bite on a multi-CCD chip. That could plausibly account for the gap in magnitude between your +170% and my +22%, on top of your library just being over 10x the size of what I tested here. If you want to check it: pin the process to one CCD (Process Lasso, or `start /affinity` with a mask covering only one CCD's cores) and rerun concurrency=8. If it moves a lot closer to flat, that's your answer.

On the two things I hadn't actually answered yet:

**Why 4** - honestly, no good reason. Went back to the commit that added it and the comment on that line: neither says why 4 specifically, only why a cap exists at all. It was a sensible-sounding starting value, not derived from anything. Given the real numbers above, more concurrency does measurably cost something now, so there's at least a case for a lower default than there was a week ago, but I don't have enough here to tell you what the right number is either.

**The per-drive queueing idea** - leaving that one for later, as you said. If you want a hand with it once you're ready to spec it out, happy to help then.

beeradmoore: > The cover-resize path is flat under Debug as well, same as Release.

(outside of scope of this PR but) TBH for further optimisation down the line the image downloading / resizing should be done independtly of the game loading. If a user is on a slow internet connection or game art CDN is slow/down then I think it will increase the time it takes to do a full game scan. Internet conneciton speed and responsiveness could be a difference between these results. I'll double check and also file an issue to put cover loading outside of the main "game is loading" check as it would prevent the game assets being able to be swapped. If this is the case it is something that should not forgotten in a PR comment section and should be added at some time.

Disk speed _could_ be a difference. But I am using a 4TB Samsung 990 Pro and a 2TB 990 Pro so if you have something faster it would have to be A LOT faster to show different numbers. Could be IO difference of my 5950X to the drives vs your hardware (like you were saying about CCD). But even then I would expect some perf difference, but not 10x difference. My 9800X3D machine has a lot less games installed but I can also test out single CCD as well.

With those hardware differences (+4 drive vs 2 which may land on better concurrency by luck without any advanced queing system) I could see yours being faster, it just does not make sense to be 2-8x faster when you have 10x as many games. That seems like a rather large gap.

I am not running any anti-virus outside of microsoft defender, and I have all of Windows security features (eg. memory integrity) turned on. If you have AV off I could imagine it could be faster.

> but what to do with any of this is entirely your call

I am more than happy to merge this in with semaphore slim limit of 4. There may be an optimal number elsewhere but I think both of our tests show that this seems to be better than letting it run wild with whatever threadpool was letting it get up to.

(EDIT: It would also be interesting to see if sysmon or something can see the number of threads max an application used)

beeradmoore: On my 9800X3D machine I have 38 games across 2 SSDs. A 4TB crucial T700 and a 2TB Samsung 990 Pro. I did the same test before, deleted `bin/`. Just before the check of `if (shouldUpdateCover == true)` I set `shouldUpdateCover = false` to prevent any cover loading.

| Test                       | Time (seconds) | shouldUpdateCover = false¹ |
|----------------------------|----------------|----------------------------|
| Before this PR             | 31.08          | 9.82                       |
| This PR - SemaphoreSlim(8) | 27.15          | 9.74                       |
| This PR - SemaphoreSlim(4) | 22.89          | 9.76                       |
| This PR - SemaphoreSlim(2) | 16.35          | 9.78                       |
| This PR - SemaphoreSlim(1) | 12.59          | 9.61                       |

¹ `shouldUpdateCover = false` was tested with a clean bin directory. However the test was ran once, not an average of three runs.

This shows me that the run to run times don't really change when we take cover updating out as a requirement of if a loading should be used. `SemaphoreSlim` is doing the heavy lifting here by reducing network or other I/O, but not specifically looking for DLL files in the filepath.

I think `SemaphoreSlim(4)` should stay in as it does have an advantage. However because this limitation is now based on image load times I would not be comfortable bumping it up/down as I am not sure if this hurts faster/slower connections more/less. I think in general this is better than before this PR so that should be given the green light. In the future I will refactor image loading out of the "Game is loading" check and let it be a separate call completely. 

I went and did a single test of `SemaphoreSlim(4)` on my 5950X machine with `shouldUpdateCover = false` and it loaded games in 5.12 seconds. What is wrong with my 9800X3D machine (😭)

beeradmoore (approved):

PR #936, opened 2026-09-07
Banner Feature

hammadk306:
## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Translation update

### DIFF
--- a/src/MainWindow.xaml
+++ b/src/MainWindow.xaml
@@ -71,5 +71,32 @@
             <TextBlock Grid.Row="1" x:Name="LoadingProgressText" VerticalAlignment="Top" Text="{x:Bind ViewModel.LoadingMessage, Mode=OneWay}" HorizontalAlignment="Center" Style="{StaticResource SubtitleTextBlockStyle}" />
 
         </Grid>
+
+        <!--
+            Persistent, low-key indicator that a new version is available.
+            Unlike the update dialog (which only appears once per version) this stays
+            visible for as long as an update is available, so it can't be missed if the
+            dialog was dismissed. Clicking it re-opens the update dialog.
+        -->
+        <Button x:Name="UpdateAvailableButton"
+                Grid.Row="1"
+                HorizontalAlignment="Right"
+                VerticalAlignment="Bottom"
+                Margin="0,0,16,16"
+                Padding="14,8"
+                CornerRadius="999"
+                BorderThickness="1"
+                Background="{ThemeResource SystemFillColorSuccessBackgroundBrush}"
+                BorderBrush="{ThemeResource SystemFillColorSuccessBrush}"
+                Visibility="{x:Bind ViewModel.UpdateAvailableVisibility, Mode=OneWay}"
+                Click="UpdateAvailableButton_Click">
+            <StackPanel Orientation="Horizontal" Spacing="8">
+                <Ellipse Width="8" Height="8" Fill="{ThemeResource SystemFillColorSuccessBrush}" VerticalAlignment="Center" />
+                <TextBlock Text="{x:Bind ViewModel.UpdateAvailableText, Mode=OneWay}"
+                           Foreground="{ThemeResource SystemFillColorSuccessBrush}"
+                           FontSize="12"
+                           VerticalAlignment="Center" />
+            </StackPanel>
+        </Button>
     </Grid>
 </Window>
--- a/src/MainWindow.xaml.cs
+++ b/src/MainWindow.xaml.cs
@@ -178,6 +178,9 @@ public sealed partial class MainWindow : Window
     LibraryPage? libraryPage;
     SettingsPage? settingsPage;
 
+    Data.GitHub.GitHubUpdater? _gitHubUpdater;
+    Data.GitHub.GitHubRelease? _availableGitHubRelease;
+
     public GameGridPage? GameGridPage => gameGridPage;
 
     void GoToPage(string page)
@@ -258,7 +261,7 @@ public sealed partial class MainWindow : Window
         }
         */
 
-        var gitHubUpdater = new Data.GitHub.GitHubUpdater();
+        var gitHubUpdater = _gitHubUpdater ??= new Data.GitHub.GitHubUpdater();
 
         // If this is a GitHub build check if there is a new version.
         var newUpdateTask = gitHubUpdater.CheckForNewGitHubRelease(false);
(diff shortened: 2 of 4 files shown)


beeradmoore:
Hey @hammadk306 , I tested this and I think there could be some changes. If you don't have time to do these or it is too complicated please let me know and I'll just build off what you have here.

src/Translations/en-US/Resources.resw
```diff
@@ -688,6 +688,10 @@ If you have checked these and your game is still not showing up there may be a b
   <data name="GitHubUpdater_UpdateAvailable" xml:space="preserve">
     <value>Update Available</value>
   </data>
+  <data name="MainWindow_UpdateAvailableBannerTemplate" xml:space="preserve">
+    <value>Update available: {0}</value>
```
beeradmoore: Based on the new update style I think this should be re-worded because "Update" is the action button. Either "Update" isn't the action button or this should not say the word update. It could say "New version available". 

Because of the `InfoBar` it could also put that in the title and version number as the text. Or both as the text, depends what looks good.

src/MainWindow.xaml
```diff
@@ -71,5 +71,32 @@
...
 
         </Grid>
+
+        <!--
+            Persistent, low-key indicator that a new version is available.
+            Unlike the update dialog (which only appears once per version) this stays
+            visible for as long as an update is available, so it can't be missed if the
+            dialog was dismissed. Clicking it re-opens the update dialog.
+        -->
+        <Button x:Name="UpdateAvailableButton"
```
beeradmoore: I think this should be an `InfoBar` with an action button instead of a button.

InfoBar can have title, text, and then button. I think this could be changed text to what you had above and then the button is "Update" or something.

Image

Check out WinUI3 gallery to see InfoBar in action. 
https://apps.microsoft.com/detail/9P3JFPWWDZRC?hl=en-us&gl=AU&ocid=pdpshare

There is an issue with infobar where the background is transparent, if you find the method `ShowTempInfoBar` you will find a comment I had left about it. I assume we can use the same kind of fix somehow.

src/MainWindow.xaml
```diff
@@ -71,5 +71,32 @@
...
+                HorizontalAlignment="Right"
+                VerticalAlignment="Bottom"
+                Margin="0,0,16,16"
+                Padding="14,8"
+                CornerRadius="999"
+                BorderThickness="1"
+                Background="{ThemeResource SystemFillColorSuccessBackgroundBrush}"
+                BorderBrush="{ThemeResource SystemFillColorSuccessBrush}"
+                Visibility="{x:Bind ViewModel.UpdateAvailableVisibility, Mode=OneWay}"
+                Click="UpdateAvailableButton_Click">
```
beeradmoore: For `Click` event I'd rather push this into the ViewModel. Look up `[RelayCommand]` in the codebase and you will find some methods in the models. 

```csharp
[RelayCommand]
async Task DoThingAsync()
{
...
}
```

This will use codegen and make a command called `DoThingCommand` which you can then use in the xaml with a binding to `Command` instead of `Click`.

src/MainWindow.xaml
```diff
@@ -71,5 +71,32 @@
...
+                Grid.Row="1"
+                HorizontalAlignment="Right"
+                VerticalAlignment="Bottom"
+                Margin="0,0,16,16"
+                Padding="14,8"
+                CornerRadius="999"
+                BorderThickness="1"
+                Background="{ThemeResource SystemFillColorSuccessBackgroundBrush}"
+                BorderBrush="{ThemeResource SystemFillColorSuccessBrush}"
+                Visibility="{x:Bind ViewModel.UpdateAvailableVisibility, Mode=OneWay}"
```
beeradmoore: I think with an infobar we can use its `IsOpen` property instead of `IsVisible`.

src/MainWindow.xaml.cs
```diff
@@ -178,6 +178,9 @@ void MainNavigationView_ItemInvoked(NavigationView sender, NavigationViewItemInv
     LibraryPage? libraryPage;
     SettingsPage? settingsPage;
 
+    Data.GitHub.GitHubUpdater? _gitHubUpdater;
```
beeradmoore: I think these values can also move back to the model. This page is probably a bad example to work off because it comes before I started more heavily using MVVM and it does not use it as heavily as I would otherwise want it to.

src/MainWindow.xaml.cs
```diff
@@ -376,13 +379,33 @@ async void MainNavigationView_Loaded(object sender, RoutedEventArgs e)
         await newUpdateTask;
         if (newUpdateTask.Result is not null)
         {
+            _availableGitHubRelease = newUpdateTask.Result;
+
+            // Show a persistent, low-key indicator so the user can't miss that an
+            // update is available even if they dismiss (or never see) the dialog below.
+            ViewModel.UpdateAvailableText = ResourceHelper.GetFormattedResourceTemplate("MainWindow_UpdateAvailableBannerTemplate", newUpdateTask.Result.Name);
+            ViewModel.UpdateAvailableVisibility = Visibility.Visible;
```
beeradmoore: If you wanted to get fancy you can make `Data.GitHub.GitHubRelease? _availableGitHubRelease;` an observable property in the model.

`UpdateAvailableText` can just be `public string UpdateAvailableText  => (AvailableGitHubRelease is null) ? string.Empty : ResourceHelper.GetFormattedResourceTemplate("MainWindow_UpdateAvailableBannerTemplate", AvailableGitHubRelease.Name);` so that way changing `AvailableGitHubRelease` will cause this text to update. 

Then on the observable property for `AvailableGitHubRelease` you can add `[NotifyPropertyChangedFor(nameof(UpdateAvailableText)]` (there are some other examples of this in the codebase). You can do the same to set the `IsInfoBarUpdateOpen` property you can also make. That way if `AvailableGitHubRelease` is null like it is by default nothing shows. You set it to a new `GitHubRelease` object and it will cause those other two properties to update automatically.

src/MainWindow.xaml.cs
```diff
@@ -376,13 +379,33 @@ async void MainNavigationView_Loaded(object sender, RoutedEventArgs e)
...
+            // The dialog itself is still only auto-shown once per version so we don't
+            // nag the user with a popup every single time they open the app.
             if (gitHubUpdater.HasPromptedBefore(newUpdateTask.Result) == false)
             {
                 await gitHubUpdater.DisplayNewUpdateDialog(newUpdateTask.Result, MainNavigationView.XamlRoot);
             }
         }
     }
 
+    async void UpdateAvailableButton_Click(object sender, RoutedEventArgs e)
```
beeradmoore: As above, this can move to the model.

You may have to update the model to take a reference of `MainWindow` so you can reference it in order to access `XamlRoot` for the dialogs.

beeradmoore (approved):

src/Translations/en-US/Resources.resw
beeradmoore: I clicked the wrong button, all these comments are meant to be changes requested

beeradmoore (changes requested):

src/MainWindow.xaml
hammadk306: It's done

src/MainWindow.xaml
hammadk306: Done

hammadk306 (commented):
The requested updates/changes are done.

src/MainWindow.xaml
```diff
@@ -71,5 +71,30 @@
...
+    Severity="Informational"
+    Title="New version available"
+    Message="{x:Bind ViewModel.UpdateAvailableText, Mode=OneWay}"
+    IsClosable="False">
+
+    <InfoBar.Action>
+        <Button
+            Content="Update"
+            Command="{x:Bind ViewModel.UpdateAvailableCommand}"
+            CommandParameter="{x:Bind MainNavigationView.XamlRoot}" />
```
beeradmoore, line 95: Using MVVM can sometimes be annoying to access this. I had never thought of passing it through `CommandParameter` before. Neat!

hammadk306: Thanks! The goal was to avoid having the ViewModel know anything about the view while still being able to use APIs that require a `XamlRoot`. Passing it through `CommandParameter` felt like a reasonable compromise and keeps the command reusable. Glad it sparked an idea!

src/MainWindow.xaml.cs
```diff
@@ -1,412 +1,78 @@
...
-using Windows.System;
+using System.Threading.Tasks;
 
 namespace DLSS_Swapper;
 
-/// <summary>
-/// An empty window that can be used on its own or navigated to within a Frame.
-/// </summary>
-public sealed partial class MainWindow : Window
+public partial class MainWindowModel : ObservableObject
```
beeradmoore, line 10: I think something broke. `MainWindowModel` was moved into `MainWindow` and now `MainWindow` class no longer exists.

beeradmoore (changes requested):
Hey @hammadk306 , I think something broke. See new comment below.

PR #938, opened 2026-09-08
feat: integrate generic DLSS 5 Autopilot support

YoshKoz:
## Description of Changes

Adds a generic per-game DLSS 5 integration in the game detail view. It downloads the latest DLSS5-Autopilot Windows release from its official GitHub repository, verifies the published SHA-256 checksum, and invokes the tool against the selected game folder.

The integration provides install/update and remove actions for every detected game, so future games do not require title-specific code. It keeps this workflow separate from ordinary DLL swapping and does not bundle NVIDIA, OptiScaler, or ReShade binaries.

The external tool selects the appropriate route and handles game-specific files and rollback. The UI warns that the integration is unofficial and should not be used with multiplayer or anti-cheat games.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Translation update

## Issues Resolved

None.

## Verification

- `dotnet restore "DLSS Swapper.sln"`
- `dotnet build "DLSS Swapper.sln" --configuration Debug --no-restore -p:Platform=x64`
- Build completed with 0 warnings and 0 errors.

Open questions for review: preferred upstream policy for external mod providers, localization coverage, and whether a separate settings toggle or provider abstraction is required before merge.

### DIFF
--- a/src/Data/DLSS5/DLSS5AutopilotManager.cs
+++ b/src/Data/DLSS5/DLSS5AutopilotManager.cs
@@ -1,0 +1,183 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Cryptography;
+using System.Text.Json;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace DLSS_Swapper.Data.DLSS5;
+
+internal sealed class DLSS5AutopilotManager
+{
+    const string Repository = "Kizzuwatnaa/DLSS5-Autopilot";
+    static readonly HttpClient HttpClient = CreateHttpClient();
+    static readonly SemaphoreSlim InstallLock = new SemaphoreSlim(1, 1);
+
+    static HttpClient CreateHttpClient()
+    {
+        var client = new HttpClient();
+        client.DefaultRequestHeaders.UserAgent.ParseAdd("DLSS-Swapper-DLSS5-Integration");
+        return client;
+    }
+
+    public static bool IsInstalled(string gamePath)
+    {
+        try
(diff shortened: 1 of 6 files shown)


beeradmoore:
Hey @YoshKoz , thanks for opening a PR.

As it is at the moment I do not want to merge this in. Mainly because DLSS Swapper is not a tool to download mods, especially those from external sources that are out of my control. I would not want an external account to be breached and have the users here be victims to malware because they updated a tool that this app said they should update.

There may be a possibility to add support for a user to go to settings and point to a folder where they have a downloaded copy and then the rest of the integration can run as intended. This flow would be true for NVPI or other tools. I would still have have to run the PR and see how that is integrated currently. 

There is an additional issue where adding support for this means for the rest of the existance of the app I will have to be providing support for someone elses mod. Even though that support is "please go report this to XYZ" it is quite time consuming. I know this for a fact because "DLSS 5 Swapper" has taken up a large portion of my free time sending people elsewhere. And it isn't even integrated into this app 😂

YoshKoz:
No worries, I completely understand the concerns around downloading and supporting external mods.

I was wondering though: would you be open to adding support for DLSS 5 Neural Rendering itself in DLSS Swapper instead? That seems like it would fit the existing scope of the app much better, without depending on external mod providers or tools.

beeradmoore:
Thanks for understanding. I never want to come in and tell people no after they have spent time to make something for the good of the app.

Support for DLSS NR is done in my streamline branch which also contains support for Streamline, DirectStorage, and FSR4. But much like how it currently works our intent is only to update existing DLLs, not to inject new functionality. 

However if you have a mod that adds nvngx_dlssnr.dll in the game then DLSS Swapper can update the DLL (and therefor AI model) that the mod would be using

